PR #24, opened 2021-08-08
[BUG] Use correct value for trace gas

patrick-ogrady:
### Changes
* Return correct `gasUsed` in `trace`
* Return hex values in metadata to support large numbers


aaronbuchwald (approved):
LGTM

PR #50, opened 2021-11-06
ERC-20/ERC-721 Support

learyce:


cache/lru_cache.go
```diff
@@ -0,0 +1,140 @@
+// (c) 2019-2020, Ava Labs, Inc. All rights reserved.
```
learyce, line 1: obviously a little copy/pasta from avalanchego, but I don't want to couple the cache implementation between the two.

patrick-ogrady: IMO it is totally ok to import this direct (there are already MANY coupled dependencies to AvalancheGo).

mapper/types.go
```diff
@@ -43,6 +43,16 @@ var (
 		Decimals: 18, //nolint:gomnd
 	}
 
+	Erc721DefaultCurrency = &types.Currency{
+		Symbol:   "ERC721",
```
patrick-ogrady, line 47: I think you want to pull this from the contract instead of specifying `ERC721` here. You should put that in the metadata as "Type"

client/contract.go
```diff
@@ -0,0 +1,69 @@
...
+
+	token, err := NewContractInfoToken(contractAddress, c.ethClient)
+	if err != nil {
+		return nil, err
+	}
+	symbol, symbolErr := token.Symbol(nil)
+	decimals, decimalErr := token.Decimals(nil)
+
+	// Any of these indicate a failure to get complete information from contract
+	if symbolErr != nil || decimalErr != nil || symbol == "" || decimals == 0 {
```
patrick-ogrady, line 55: IMO, if we can't populate the symbol or decimals, we should not index the transfer.

service/config.go
```diff
@@ -9,11 +9,13 @@ import (
...
-	NetworkID        *types.NetworkIdentifier
-	GenesisBlockHash string
-	AvaxAssetID      string
+	Mode                 string
+	ChainID              *big.Int
+	NetworkID            *types.NetworkIdentifier
+	GenesisBlockHash     string
+	AvaxAssetID          string
+	IndexErc721Transfers bool
+	IndexDefaultTokens   bool
```
patrick-ogrady, line 18: Why not call this `IndexErc20Transfers`?

client/contract.go
learyce: We handle the indexing downstream of this, we'll still want to add an entry in the cache so that we can avoid un-necessary rpc calls

client/contract.go
```diff
@@ -0,0 +1,70 @@
...
+	"fmt"
+	"strings"
+
+	"github.com/ava-labs/avalanchego/cache"
+	"github.com/ethereum/go-ethereum/common"
+	"github.com/ethereum/go-ethereum/ethclient"
+)
+
+const (
+	contractCacheSize = 1000
```
patrick-ogrady, line 13: nit: base2 cache sizes

cmd/server/config.go
```diff
@@ -15,13 +15,15 @@ var (
...
-	GenesisBlockHash string `json:"genesis_block_hash"`
+	Mode                 string `json:"mode"`
+	RPCEndpoint          string `json:"rpc_endpoint"`
+	ListenAddr           string `json:"listen_addr"`
+	NetworkName          string `json:"network_name"`
+	ChainID              int64  `json:"chain_id"`
+	LogRequests          bool   `json:"log_requests"`
+	GenesisBlockHash     string `json:"genesis_block_hash"`
+	IndexErc721Transfers bool   `json:"index_erc721_transfers"`
+	IndexDefaultTokens   bool   `json:"index_default_tokens"`
```
patrick-ogrady, line 26: nit: `IndexErc20Transfers` not `IndexDefaultTokens`...also put it above 721

mapper/transaction.go
```diff
@@ -63,6 +72,86 @@ func Transaction(
...
+			receiptOp := types.Operation{
+				OperationIdentifier: &types.OperationIdentifier{
+					Index: int64(len(ops)),
+				},
+				Status:  types.String(StatusSuccess),
+				Type:    OpErc721Transfer,
+				Amount:  Erc721Amount(erc721Index, contractAddress, contractInfo.Symbol, contractInfo.Decimals, false),
+				Account: Account(ConvertHashToAddress(&addressTo)),
+			}
+			sendingOp := types.Operation{
```
patrick-ogrady, line 106: Nit: you are going to want to populate the `related_operations` field for both 20 and 721

mapper/address.go
```diff
@@ -0,0 +1,14 @@
...
+
+import (
+	ethcommon "github.com/ethereum/go-ethereum/common"
+)
+
+// ConvertHashToAddress uses the last 20 bytes of a common.Hash to create a common.Address
+func ConvertHashToAddress(hash *ethcommon.Hash) *ethcommon.Address {
+	if hash == nil {
+		return nil
+	}
```
patrick-ogrady, line 11: You may want to ensure len(hash) == 32 (as a constant) to avoid panic

mapper/types.go
```diff
@@ -11,17 +11,23 @@ const (
...
-	OpSelfDestruct = "SELFDESTRUCT"
-	OpCallCode     = "CALLCODE"
-	OpDelegateCall = "DELEGATECALL"
-	OpStaticCall   = "STATICCALL"
-	OpDestruct     = "DESTRUCT"
-	OpImport       = "IMPORT"
-	OpExport       = "EXPORT"
+	TokenTypeMetadata       = "tokenType"
+	ContractAddressMetadata = "contractAddress"
+	IndexTransferedMetadata = "indexTransfered"
```
patrick-ogrady, line 16: You should put `Unknown *` here too IMO

service/config.go
patrick-ogrady: I still think we should do IndexErc20Transfers. Commented again above.

service/service_account.go
```diff
@@ -64,14 +63,59 @@ func (s AccountService) AccountBalance(
...
+
+		amount := mapper.Erc20Amount(response, contractAddress, contractInfo.Symbol, contractInfo.Decimals, false)
+		balances = append(balances, amount)
+
+		if err != nil {
+			return nil, wrapError(errInternalError, err)
+		}
+	}
+
+	balances = append(balances, mapper.AvaxAmount(avaxBalance))
```
patrick-ogrady, line 111: Add the avaxBalance as the first element:
```
balances := []*types.Balances{avaxBalance}
```

service/service_account.go
```diff
@@ -64,14 +63,59 @@ func (s AccountService) AccountBalance(
...
 	}
 
+	avaxBalance, balanceErr := s.client.BalanceAt(context.Background(), address, header.Number)
+	if balanceErr != nil {
+		return nil, wrapError(errInternalError, balanceErr)
+	}
+
+	var balances []*types.Amount
+
+	for _, currency := range req.Currencies {
```
patrick-ogrady, line 73: Nit: you may want to skip this if `IndexErc20Transfers` is false

service/types.go
```diff
@@ -286,3 +288,8 @@ func (m *accountMetadata) UnmarshalJSON(data []byte) error {
 	m.Nonce = nonce
 	return nil
 }
+
+// has0xPrefix validates str begins with '0x' or '0X'.
+func has0xPrefix(str string) bool {
```
patrick-ogrady, line 293: Nit: is there not a helper already in eth to do this? if not, you should use `strings.HasPrefix`

service/types.go
```diff
@@ -101,6 +102,7 @@ type metadata struct {
 	Nonce    uint64   `json:"nonce"`
 	GasPrice *big.Int `json:"gas_price"`
 	GasLimit uint64   `json:"gas_limit"`
+	Data     string   `json:"tx_data"`
```
patrick-ogrady, line 105: I would highly recommend not touching anything in construction on this pass.

service/service_construction.go
```diff
@@ -397,7 +397,7 @@ func (s ConstructionService) ConstructionPayloads(
 	gasPrice := metadata.GasPrice
 	gasLimit := metadata.GasLimit
 	chainID := s.config.ChainID
-	transferData := []byte{}
+	transferData := []byte(metadata.Data)
```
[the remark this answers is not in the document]
patrick-ogrady: I would highly recommend not touching anything in construction on this pass.

client/contract.go
```diff
@@ -0,0 +1,70 @@
...
+	if symbolErr != nil || decimalErr != nil || symbol == "" || decimals == 0 {
+		if isErc20 {
+			symbol = UnknownERC20Symbol
+			decimals = UnknownERC20Decimals
+		} else {
+			symbol = UnknownERC721Symbol
+			decimals = UnknownERC721Decimals
+		}
+	}
+	contractInfo := ContractInfo{Symbol: symbol, Decimals: decimals}
```
patrick-ogrady, line 65: Use `&ContractInfo` here...then you will only keep one copy of `ContractInfo` in memory.

client/contract.go
```diff
@@ -0,0 +1,70 @@
...
+		cache:     cache,
+	}, nil
+}
+
+// ContractInfo returns the ContractInfo for a specific address
+func (c *ContractClient) ContractInfo(contractAddress common.Address, isErc20 bool) (*ContractInfo, error) {
+	cachedInfo, isCached := c.cache.Get(contractAddress)
+
+	if isCached {
+		castCachedInfo := cachedInfo.(ContractInfo)
```
patrick-ogrady, line 44: You'll need to cast this as `(*ContractInfo)` after making the pointer change below

client/evm_logs.go
```diff
@@ -0,0 +1,72 @@
...
+// NewEvmLogsClient returns a new EVM Logs client
+func NewEvmLogsClient(endpoint string) (*EvmLogsClient, error) {
+	endpoint = strings.TrimSuffix(endpoint, "/")
+
+	c, err := ethclient.Dial(fmt.Sprintf("%s%s", endpoint, prefixEth))
+	if err != nil {
+		return nil, err
+	}
+
+	cache := &cache.LRU{Size: logCacheSize}
```
patrick-ogrady, line 35: I'd suggest removing this cache. During normal operation, you'll never fetch the same block twice.

client/client.go
```diff
@@ -53,8 +59,20 @@ func NewClient(endpoint string) (Client, error) {
 		return nil, err
 	}
 
+	evmlogs, err := NewEvmLogsClient(endpoint)
```
patrick-ogrady, line 62: Why not reuse the raw `ethclient` returned by `NewEthClient`?

cmd/server/config.go
```diff
@@ -6,22 +6,29 @@ import (
...
-	LogRequests      bool   `json:"log_requests"`
-	GenesisBlockHash string `json:"genesis_block_hash"`
+	Mode                       string   `json:"mode"`
+	RPCEndpoint                string   `json:"rpc_endpoint"`
+	ListenAddr                 string   `json:"listen_addr"`
+	NetworkName                string   `json:"network_name"`
+	ChainID                    int64    `json:"chain_id"`
+	LogRequests                bool     `json:"log_requests"`
+	GenesisBlockHash           string   `json:"genesis_block_hash"`
+	IngestionMode              string   `json:"ingestion_mode"`
```
patrick-ogrady, line 28: I'd add a `\n` after `GenesisBlockHash` and put `IngestionMode`/`TokenWhitelist` with it

cmd/server/config.go
```diff
@@ -6,22 +6,29 @@ import (
...
+	StandardModeTokenWhitelist []string `json:"standard_token_addresses"`
```
patrick-ogrady, line 29: nit: TokenWhitelist

cmd/server/config.go
```diff
@@ -6,22 +6,29 @@ import (
...
+
+	IndexDefaultTokens bool `json:"index_default_tokens"`
```
patrick-ogrady, line 31: I recommend using more informative naming here. I think this is `IndexUnknownTokens`?

mapper/address.go
```diff
@@ -0,0 +1,14 @@
+package mapper
+
+import (
+	ethcommon "github.com/ethereum/go-ethereum/common"
+)
+
+// ConvertHashToAddress uses the last 20 bytes of a common.Hash to create a common.Address
```
patrick-ogrady, line 7: I'd recommend elaborating on what type of hash you are parsing.

mapper/address.go
```diff
@@ -0,0 +1,14 @@
...
+func ConvertHashToAddress(hash *ethcommon.Hash) *ethcommon.Address {
+	if hash == nil {
+		return nil
+	}
+	address := ethcommon.BytesToAddress(hash[12:32])
```
patrick-ogrady, line 12: Do you just accept panic here? or can you return an error?

mapper/amount.go
```diff
@@ -27,3 +28,50 @@ func FeeAmount(value int64) *types.Amount {
...
+	contractSymbol string,
+	contractDecimal uint8,
+	isSender bool) *types.Amount {
+	index := indexHash.Big()
+	metadata := make(map[string]interface{})
+	metadata[TokenTypeMetadata] = "ERC721"
+	metadata[ContractAddressMetadata] = contractAddress.String()
+	metadata[IndexTransferedMetadata] = index.String()
+
+	return &types.Amount{
```
patrick-ogrady, line 44: IMO this doesn't make sense to have an amount. You should just return metadata.

mapper/transaction.go
```diff
@@ -62,7 +73,75 @@ func Transaction(
 
 	traceOps := traceOps(flattenedTrace, len(feeOps))
 	ops = append(ops, traceOps...)
+	for _, transferLog := range transferLogs {
```
patrick-ogrady, line 76: We shouldn't even be fetching logs in standard mode (as it will add latency).

mapper/transaction.go
patrick-ogrady: This comment still stands?

mapper/transaction.go
```diff
@@ -62,7 +73,75 @@ func Transaction(
...
+				Account:  Account(ConvertHashToAddress(&addressFrom)),
+				Metadata: metadata,
+				RelatedOperations: []*types.OperationIdentifier{
+					{
+						Index: int64(len(ops)),
+					},
+				},
+			}
+
+			ops = append(ops, &receiptOp)
```
patrick-ogrady, line 128: It is more canonical in rosetta to do send before receive.

mapper/transaction.go
```diff
@@ -400,3 +479,69 @@ func traceOps(trace []*client.FlatCall, startIndex int) []*types.Operation {
...
 	return ops
 }
+
+func parseErc20Txs(transferLog ethtypes.Log, contractInfo *clientTypes.ContractInfo, opsLen int64) []*types.Operation {
+	ops := []*types.Operation{}
+
+	contractAddress := transferLog.Address
+	addressFrom := transferLog.Topics[1]
+	addressTo := transferLog.Topics[2]
+	log.Println(addressFrom.Hex())
```
patrick-ogrady, line 489: remove logging

mapper/transaction.go
```diff
@@ -400,3 +479,69 @@ func traceOps(trace []*client.FlatCall, startIndex int) []*types.Operation {
...
+		Type:    OpErc20Transfer,
+		Amount:  Erc20Amount(transferLog.Data, contractAddress, contractInfo.Symbol, contractInfo.Decimals, true),
+		Account: Account(ConvertHashToAddress(&addressFrom)),
+		RelatedOperations: []*types.OperationIdentifier{
+			{
+				Index: opsLen,
+			},
+		},
+	}
+	ops = append(ops, &receiptOp)
```
patrick-ogrady, line 543: send before receive

mapper/transaction.go
```diff
@@ -62,7 +73,75 @@ func Transaction(
...
+			addressFrom := transferLog.Topics[1]
+			addressTo := transferLog.Topics[2]
+			erc721Index := transferLog.Topics[3] // Erc721 4th topic is the index.  Data is empty
+			metadata := make(map[string]interface{})
+
+			metadata[TokenTypeMetadata] = "ERC721"
+			metadata[ContractAddressMetadata] = contractAddress.String()
+			metadata[IndexTransferedMetadata] = erc721Index.String()
+
+			receiptOp := types.Operation{
```
patrick-ogrady, line 104: You may want to do the same zeroAdr check you did for ERC-20 for mint/burn?

rosetta-cli-conf/testnet/config.json
```diff
@@ -40,7 +40,6 @@
     "bootstrap_balances": "",
     "interesting_accounts": "",
     "reconciliation_disabled": false,
-    "inactive_discrepency_search_disabled": false,
```
patrick-ogrady, line 43: Why remove this line?

service/service_account.go
```diff
@@ -64,14 +64,65 @@ func (s AccountService) AccountBalance(
...
+	if balanceErr != nil {
+		return nil, wrapError(errInternalError, balanceErr)
+	}
+
+	var balances []*types.Amount
+	balances = append(balances, mapper.AvaxAmount(avaxBalance))
+	for _, currency := range req.Currencies {
+		value, ok := currency.Metadata[mapper.ContractAddressMetadata]
+		if !ok {
+			if currency.Decimals == 18 && strings.ToLower(currency.Symbol) == "avax" {
```
patrick-ogrady, line 77: You should use consts to perform this check

service/service_account.go
```diff
@@ -64,14 +64,65 @@ func (s AccountService) AccountBalance(
...
+		}
+
+		contractInfo, err := s.client.ContractInfo(contractAddress, true)
+		var amount *types.Amount
+		if err != nil {
+			return nil, wrapError(errInternalError, err)
+		}
+
+		if contractInfo.Symbol == client.UnknownERC20Symbol {
+			amount = mapper.Erc20Amount(response, contractAddress, client.UnknownERC20Symbol, client.UnknownERC20Decimals, false)
```
patrick-ogrady, line 112: Don't we always have the symbol/decimals from the request? I think it is ok if we don't verify them here if the balance call works.

mapper/transaction.go
patrick-ogrady: nvm saw this is handled below...may make sense to put a comment around this.

service/types.go
patrick-ogrady: This point still stands.

service/types.go
```diff
@@ -286,3 +287,8 @@ func (m *accountMetadata) UnmarshalJSON(data []byte) error {
 	m.Nonce = nonce
 	return nil
 }
+
+// has0xPrefix validates str begins with '0x' or '0X'.
+func has0xPrefix(str string) bool {
```
patrick-ogrady, line 292: Please note this comes from go-ethereum: https://github.com/ethereum/go-ethereum/blob/0ba0b81e54483978ab823665afb627ab4999db06/common/hexutil/hexutil.go#L189-L191

mocks/client/client.go
```diff
@@ -455,3 +455,68 @@ func (_m *Client) TxPoolStatus(_a0 context.Context) (*client.TxPoolStatus, error
...
+	if rf, ok := ret.Get(1).(func(context.Context, common.Hash, common.Hash) error); ok {
+		r1 = rf(_a0, _a1, _a2)
+	} else {
+		r1 = ret.Error(1)
+	}
+
+	return r0, r1
+}
+
+func (_m *Client) CallContract(_a0 context.Context, _a1 interfaces.CallMsg, _a2 *big.Int) ([]byte, error) {
```
patrick-ogrady, line 502: Would be nice to add some testing for these

client/evm_logs.go
learyce: This cache is used a lot.  The cleanest way to get EVM logs was to fetch them on the transaction level.  So each transaction within a block will pull the logs from the cache

client/client.go
learyce: Wanted a custom implementation with caching and also handled filtering so only logs relevant to a specific transaction hash are returned

mapper/address.go
learyce: Shouldn't be an error or panic, ethcommon.Hash is always 32 bytes long

learyce: If it changes without us knowing, I'm fine with the panic

mapper/amount.go
learyce: Unused function from before our discussion.  Deleting

mapper/transaction.go
learyce: Yep, we shouldn't have any logs.  Originally had it more explicit with an if statement, but the linter said my code was too complex :laughing: and I had to remove it

learyce: Also, in standard mode we will have logs if there's a non-empty token whitelist

rosetta-cli-conf/testnet/config.json
learyce: It's no longer supported in rosetta-cli

service/types.go
learyce: Don't know if my comment went through last time, but this function is not publicly accessible from go-ethereum.  That's why I reproduced here

mocks/client/client.go
learyce: yes, definitely will revisit during the construction api portion

cmd/server/config.go
```diff
@@ -66,5 +77,17 @@ func (c *config) Validate() error {
 		return errGenesisBlockRequired
 	}
 
+	if len(c.TokenWhiteList) != 0 {
+		for _, token := range c.TokenWhiteList {
+			if !ethcommon.IsHexAddress(token) {
+				return errInvalidTokenAddress
```
patrick-ogrady, line 83: Do we want to lookup the list using the contract client to make sure they are all defined?

mapper/transaction.go
```diff
@@ -400,3 +448,135 @@ func traceOps(trace []*client.FlatCall, startIndex int) []*types.Operation {
...
+	}
+	sendingOp := types.Operation{
+		OperationIdentifier: &types.OperationIdentifier{
+			Index: opsLen + 1,
+		},
+		Status:  types.String(StatusSuccess),
+		Type:    OpErc20Transfer,
+		Amount:  Erc20Amount(transferLog.Data, contractAddress, contractInfo.Symbol, contractInfo.Decimals, true),
+		Account: Account(ConvertEVMTopicHashToAddress(&addressFrom)),
+		RelatedOperations: []*types.OperationIdentifier{
```
patrick-ogrady, line 504: The related operation should be on the second op (receiptOp).

mapper/transaction.go
```diff
@@ -400,3 +448,135 @@ func traceOps(trace []*client.FlatCall, startIndex int) []*types.Operation {
...
+			{
+				Index: opsLen,
+			},
+		},
+	}
+	ops = append(ops, &sendingOp)
```
patrick-ogrady, line 510: The indexing is out of order if you just switch here.

mapper/transaction.go
```diff
@@ -400,3 +448,135 @@ func traceOps(trace []*client.FlatCall, startIndex int) []*types.Operation {
...
+		Type:     OpErc721TransferSender,
+		Account:  Account(ConvertEVMTopicHashToAddress(&addressFrom)),
+		Metadata: metadata,
+		RelatedOperations: []*types.OperationIdentifier{
+			{
+				Index: opsLen,
+			},
+		},
+	}
+	ops = append(ops, &sendingOp)
```
patrick-ogrady, line 579: see notes above on this.

mapper/transaction.go
```diff
@@ -400,3 +448,135 @@ func traceOps(trace []*client.FlatCall, startIndex int) []*types.Operation {
...
+	metadata[ContractAddressMetadata] = contractAddress.String()
+	metadata[IndexTransferedMetadata] = erc721Index.String()
+
+	if addressFrom.Hex() == zeroAddress {
+		mintOp := types.Operation{
+			OperationIdentifier: &types.OperationIdentifier{
+				Index: opsLen,
+			},
+			Status:   types.String(StatusSuccess),
+			Type:     OpErc721Mint,
```
patrick-ogrady, line 534: ❤️ ❤️

mapper/types.go
```diff
@@ -13,17 +13,29 @@ const (
...
+	OpDelegateCall  = "DELEGATECALL"
+	OpStaticCall    = "STATICCALL"
+	OpDestruct      = "DESTRUCT"
+	OpImport        = "IMPORT"
+	OpExport        = "EXPORT"
+	OpErc20Transfer = "ERC20_TRANSFER"
+	OpErc20Mint     = "ERC20_MINT"
+	OpErc20Burn     = "ERC20_BURN"
+
+	OpErc721TransferSender   = "ERC721_SENDER"
```
patrick-ogrady, line 35: nit: Send/receive not sender/receipient

rosetta-cli-conf/testnet/config.json
patrick-ogrady: wuhhhhh. that was hard to make

service/service_account.go
```diff
@@ -64,14 +64,55 @@ func (s AccountService) AccountBalance(
 		return nil, wrapError(errInternalError, metadataErr)
 	}
 
+	avaxBalance, balanceErr := s.client.BalanceAt(context.Background(), address, header.Number)
+	if balanceErr != nil {
+		return nil, wrapError(errInternalError, balanceErr)
+	}
+
+	var balances []*types.Amount
+	balances = append(balances, mapper.AvaxAmount(avaxBalance))
```
patrick-ogrady, line 73: You should only add `avaxBalance` here if `len(req.Currencies)` == 0, otherwise, you should only populate it if it is in `req.Currencies`.

service/service_account.go
```diff
@@ -64,14 +64,55 @@ func (s AccountService) AccountBalance(
...
+	for _, currency := range req.Currencies {
+		value, ok := currency.Metadata[mapper.ContractAddressMetadata]
+		if !ok {
+			if currency.Decimals == AvaxDecimals && strings.ToLower(currency.Symbol) == AvaxSymbolLower {
```
patrick-ogrady, line 77: This is not the right match check for a currency input.

patrick-ogrady: You should use `types.Hash` here: https://github.com/coinbase/rosetta-sdk-go/blob/2868f1040ba0db56acedd72d2ec745d0c2c44d57/reconciler/reconciler.go#L181-L184

service/types.go
patrick-ogrady: Yep, not a comment about that we reporduced it. More about attributing.

client/client.go
patrick-ogrady: I guess what I mean is pass the `ethClient` created in `NewEthClient` into `NewEVMLogsClient` instead of creating another.

README.md
```diff
@@ -55,7 +56,8 @@ Full configuration example:
   "listen_addr": "0.0.0.0:8080",
   "network_name": "Fuji",
   "chain_id": 43113,
-  "log_requests": true
+  "log_requests": true,
+  "genesis_block_hash" :"0x31ced5b9beb7f8782b014660da0cb18cc409f121f408186886e1ca3e8eeca96b",
 }
 ```
```
patrick-ogrady, line 63: Probably want to add the new configs you created here.

cmd/server/config.go
learyce: Could go either way, theoretically as long as it's a valid address we shouldn't care.  Probably moot, but they could want to track info on a contract that doesn't have a public Symbol/Name

README.md
```diff
@@ -58,6 +58,9 @@ Full configuration example:
   "chain_id": 43113,
   "log_requests": true,
   "genesis_block_hash" :"0x31ced5b9beb7f8782b014660da0cb18cc409f121f408186886e1ca3e8eeca96b",
+  "index_unknown_tokens": false,
+  "ingestion_mode" : "standard",
+  "token_whitelist" : []
 }
```
patrick-ogrady, line 64: You should also fil in the `WHERE` box below.

service/service_account.go
```diff
@@ -70,11 +69,11 @@ func (s AccountService) AccountBalance(
 	}
 
 	var balances []*types.Amount
-	balances = append(balances, mapper.AvaxAmount(avaxBalance))
 	for _, currency := range req.Currencies {
 		value, ok := currency.Metadata[mapper.ContractAddressMetadata]
 		if !ok {
-			if currency.Decimals == AvaxDecimals && strings.ToLower(currency.Symbol) == AvaxSymbolLower {
+			if types.Hash(currency) == types.Hash(&types.Currency{Symbol: AvaxSymbol, Decimals: AvaxDecimals}) {
+				balances = append(balances, mapper.AvaxAmount(avaxBalance))
```
patrick-ogrady, line 76: You also need to add this if `req.Currencies` is `len==0`

learyce: So exactly the way it was before?

patrick-ogrady (approved):
🏆

rosetta-cli-conf/testnet/config.json
shrimalmadhur: This was a correction in spelling `inactive_discrepency_search_disabled` -> `inactive_discrepancy_search_disabled`. I am wondering how to make it more clear while upgrading the cli packages since I am planning more config changes to make that config more clear and consistent.

PR #54, opened 2021-11-21
Atomic Transaction Cleanup

patrick-ogrady:


mapper/transaction.go
```diff
@@ -88,93 +90,114 @@ func CrossChainTransactions(
...
-		mTxIDs := map[string]struct{}{}
-		for _, in := range t.ImportedInputs {
-			mTxIDs[in.TxID.String()] = struct{}{}
-		}
-		i := 0
-		txIDs := make([]string, len(mTxIDs))
-		for txID := range mTxIDs {
-			txIDs[i] = txID
-			i++
+	for _, tx := range atomicTxs {
```
patrick-ogrady, line 111: TODO: cleanup

mapper/transaction.go
```diff
@@ -172,9 +171,50 @@ func CrossChainTransactions(
...
+	ops := []*types.Operation{}
+	for _, tx := range atomicTxs {
+		txOps, err := crossChainTransaction(len(ops), avaxAssetID, tx)
+		if err != nil {
+			return nil, err
+		}
+		ops = append(ops, txOps...)
 	}
 
+	// TODO: migrate to using atomic transaction ID instead of marking as a block
```
patrick-ogrady, line 216: We may want to change this in this release, however, this would mean CB would need to re-index things.

mapper/transaction.go
```diff
@@ -88,93 +90,114 @@ func CrossChainTransactions(
...
+		// block.
+		tx := new(evm.Tx)
+		if _, err := codecManager.Unmarshal(extra, tx); err != nil {
+			return nil, err
+		}
+		atomicTxs = append(atomicTxs, tx)
+	} else {
+		if _, err := codecManager.Unmarshal(extra, &atomicTxs); err != nil {
+			return nil, fmt.Errorf("failed to unmarshal atomic tx (AP5) due to %w", err)
+		}
```
StephenButtolph, line 204: nit: rather than initializing an empty array (that is never used) I feel like it would be cleaner to do:
```
var atomicTxs []*evm.Tx
if block.Time() < AP5Activation {
	// Prior to Apricot Phase 5, there was only one atomic transaction per
	// block.
	tx := new(evm.Tx)
	if _, err := codecManager.Unmarshal(extra, tx); err != nil {
		return nil, err
	}
	atomicTxs = []*evm.Tx{tx}
} else {
	if _, err := codecManager.Unmarshal(extra, &atomicTxs); err != nil {
		return nil, fmt.Errorf("failed to unmarshal atomic tx (AP5) due to %w", err)
	}
}
```

StephenButtolph (commented):
1 nit, otherwise lgtm

mapper/transaction.go
```diff
@@ -172,9 +171,50 @@ func CrossChainTransactions(
 			idx++
 		}
 	default:
-		panic("Unsupported transaction:" + reflect.TypeOf(t).String())
+		return nil, fmt.Errorf("Unsupported transaction: %s", reflect.TypeOf(t).String())
```
dhrubabasu, line 174:
```suggestion
		return nil, fmt.Errorf("Unsupported transaction: %T", t)
```

dhrubabasu: Also need to remove the `reflect` import since it'll be no longer used after this change

mapper/types.go
```diff
@@ -5,11 +5,13 @@ import (
...
-	MainnetAssetID = "FvwEAhmxKfeiG8SnEvq42hc6whRyY3EFYAvebMqDNDGCgxN5Z"
+	MainnetChainID       = 43114
+	MainnetAssetID       = "FvwEAhmxKfeiG8SnEvq42hc6whRyY3EFYAvebMqDNDGCgxN5Z"
+	MainnetAP5Activation = 2000000000 // TODO: SET!!
 
-	FujiChainID = 43113
-	FujiAssetID = "U8iRqJoiJm8xZHAacmvYyZVwqQx6uDNtQeP3CQ6fcgQk3JqnK"
+	FujiChainID       = 43113
+	FujiAssetID       = "U8iRqJoiJm8xZHAacmvYyZVwqQx6uDNtQeP3CQ6fcgQk3JqnK"
+	FujiAP5Activation = 1637690400 // 11/23/21 6PM UTC
```
patrick-ogrady, line 14: Replace with actual time value

PR #63, opened 2021-12-19
[CPT-142] Construction api

learyce:


cmd/server/config.go
```diff
@@ -91,3 +93,19 @@ func (c *config) Validate() error {
...
+			contractInfo, err := cli.ContractInfo(ethAdress, true)
+			if err != nil {
+				return err
+			}
+			if contractInfo.Decimals == client.UnknownERC20Decimals && contractInfo.Symbol == client.UnknownERC20Symbol {
+				return errInvalidErc20Address
+			}
+		}
+	}
+	return nil
```
dhrubabasu, line 110: The invariant check should be first:
```golang
if len(c.TokenWhiteList) == 0 {
	return nil
}
```

docker/entrypoint.sh
```diff
@@ -1,62 +0,0 @@
-#!/bin/bash
```
dhrubabasu, line 1: Why was this file deleted?

service/service_construction.go
```diff
@@ -412,14 +424,35 @@ func (s ConstructionService) ConstructionPayloads(
...
 
-	tx := ethtypes.NewTransaction(
-		nonce,
-		ethcommon.HexToAddress(checkTo),
-		amount,
-		gasLimit,
-		gasPrice,
-		transferData,
-	)
+	var tx *ethtypes.Transaction
```
dhrubabasu, line 427: The assignment of `tx` could be cleaned up. The if statement should assign `transferData` and the `ethcommon.HexToAddress` call with a single assignment of `tx` after line 455.

service/service_construction.go
```diff
@@ -588,3 +613,147 @@ func (s ConstructionService) ConstructionSubmit(
...
+		Account: &parser.AccountDescription{
+			Exists: true,
+		},
+		Amount: &parser.AmountDescription{
+			Exists:   true,
+			Sign:     parser.NegativeAmountSign,
+			Currency: mapper.AvaxCurrency,
+		},
+	}
+	nativeRecieve := parser.OperationDescription{
```
dhrubabasu, line 631: nit: receive is mispelled

service/service_construction.go
```diff
@@ -588,3 +613,147 @@ func (s ConstructionService) ConstructionSubmit(
...
+	contractAddress := common.HexToAddress(contract.(string))
+	data := generateErc20TransferData(toAddress, value)
+	gasLimit, err := s.client.EstimateGas(ctx, interfaces.CallMsg{
+		From: common.HexToAddress(fromAddress),
+		To:   &contractAddress,
+		Data: data,
+	})
+	if err != nil {
+		return 0, err
+	}
```
dhrubabasu, line 717: This if statement is unnecessary, you should return `err` directly

service/service_construction.go
```diff
@@ -588,3 +613,147 @@ func (s ConstructionService) ConstructionSubmit(
...
+	to := common.HexToAddress(toAddress)
+	gasLimit, err := s.client.EstimateGas(ctx, interfaces.CallMsg{
+		From:  common.HexToAddress(fromAddress),
+		To:    &to,
+		Value: value,
+	})
+	if err != nil {
+		return 0, err
+	}
```
dhrubabasu, line 696: This if statement is unnecessary, you should return `err` directly

cmd/server/config.go
```diff
@@ -91,3 +93,19 @@ func (c *config) Validate() error {
 
 	return nil
 }
+
+func (c *config) ValidateWhitelistOnlyValidErc20s(cli client.Client) error {
+	if len(c.TokenWhiteList) != 0 {
+		for _, token := range c.TokenWhiteList {
+			ethAdress := ethcommon.HexToAddress(token)
```
dhrubabasu, line 100: nit: address is mispelled

cmd/server/config.go
```diff
@@ -91,3 +93,19 @@ func (c *config) Validate() error {
...
+			contractInfo, err := cli.ContractInfo(ethAdress, true)
+			if err != nil {
+				return err
+			}
+			if contractInfo.Decimals == client.UnknownERC20Decimals && contractInfo.Symbol == client.UnknownERC20Symbol {
+				return errInvalidErc20Address
```
dhrubabasu, line 106: we should output the erc20 address that errors

service/service_construction.go
```diff
@@ -272,23 +270,50 @@ func (s ConstructionService) ConstructionParse(
 	if !ok {
 		return nil, wrapError(errInvalidInput, fmt.Errorf("%s is not a valid address", tx.To))
 	}
+	var opMethod string
+	var currency *types.Currency
+	var value *big.Int
+	// Erc20 transfer
+	if len(tx.Data) != 0 {
+		if !mapper.EqualFoldContains(s.config.TokenWhiteList, tx.To) {
+			return nil, wrapError(errInvalidInput, "unsupported contract address associated with transaction")
```
dhrubabasu, line 279: `tx.To` should be in this error msg

dhrubabasu (changes requested):
Left some initial thoughts

docker/entrypoint.sh
patrick-ogrady: I think this had to be an accident

service/service_construction.go
```diff
@@ -89,17 +91,13 @@ func (s ConstructionService) ConstructionMetadata(
 
 	var gasLimit uint64
 	if input.GasLimit == nil {
-		if len(input.To) == 0 || input.Value == nil {
-			// We guard against malformed inputs that may have been generated using
-			// a previous version of avalanche-rosetta.
-			gasLimit = transferGasLimit
+		if input.Currency == nil || types.Hash(input.Currency) == types.Hash(mapper.AvaxCurrency) {
```
patrick-ogrady, line 94: When would currency be nil here? I guess for ERC-721?

service/service_construction.go
```diff
@@ -272,23 +270,50 @@ func (s ConstructionService) ConstructionParse(
...
+		if !mapper.EqualFoldContains(s.config.TokenWhiteList, tx.To) {
+			return nil, wrapError(errInvalidInput, "unsupported contract address associated with transaction")
+		}
+
+		toAddress, amountSent, err := parseErc20TransferData(tx.Data)
+		if err != nil {
+			return nil, wrapError(errInvalidInput, err)
+		}
+
+		contractInfo, err := s.client.ContractInfo(*toAddress, true)
```
patrick-ogrady, line 287: Parsing must be entirely offline. You should wrap the tx to sign with whatever logic is needed to parse it: https://www.rosetta-api.org/docs/construction_api_introduction.html#completely-offline

service/helper.go
```diff
@@ -15,8 +15,9 @@ import (
 )
 
 const (
-	transferGasLimit = uint64(21000) //nolint:gomnd
-	genesisTimestamp = 946713601000  // min allowable timestamp
+	nativeTransferGasLimit = uint64(21000) //nolint:gomnd
+	erc20TransferGasLimit  = uint64(42000)
```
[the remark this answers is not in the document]
learyce: Updated to 100,000.  Let me know if we should go higher

service/service_construction.go
learyce: Some of the old test cases had a nil Currency field.  If we don't have a currency provided, we assume that it's AVAX (just to make sure we backdate the new functionality)

service/service_construction.go
```diff
@@ -588,3 +613,146 @@ func (s ConstructionService) ConstructionSubmit(
 		},
 	}, nil
 }
+
+func (s ConstructionService) CreateOperationDescription() ([]*parser.OperationDescription, error) {
```
learyce, line 617: here

service/helper.go
```diff
@@ -15,8 +15,9 @@ import (
 )
 
 const (
-	transferGasLimit = uint64(21000) //nolint:gomnd
-	genesisTimestamp = 946713601000  // min allowable timestamp
+	nativeTransferGasLimit = uint64(21000) //nolint:gomnd
+	erc20TransferGasLimit  = uint64(100000)
```
learyce, line 19: @patrick-ogrady , do you think this is high enough?

cmd/server/main.go
```diff
@@ -52,11 +52,18 @@ func main() {
 		log.Fatal("config validation error:", err)
 	}
 
-	apiClient, err := client.NewClient(cfg.RPCEndpoint)
+	apiClient, err := client.NewClient(cfg.RPCEndpoint, cfg.RPCToken)
 	if err != nil {
 		log.Fatal("client init error:", err)
 	}
 
+	// Token Address Validation requires online mode
```
patrick-ogrady, line 60: ❤️

service/helper.go
patrick-ogrady: That is too low

mapper/amount.go
```diff
@@ -41,15 +47,21 @@ func Erc20Amount(
 	if isSender {
 		decimalValue = new(big.Int).Neg(decimalValue)
 	}
-	metadata := make(map[string]interface{})
-	metadata[ContractAddressMetadata] = contractAddress.String()
 
+	currency := Erc20Currency(contractSymbol, int32(contractDecimal), contractAddress.String())
```
patrick-ogrady, line 51: When would the fee ever be paid in ERC20?

service/service_construction.go
patrick-ogrady: Agreed ^^

service/service_construction.go
patrick-ogrady: Good find

service/service_construction.go
```diff
@@ -194,8 +192,15 @@ func (s ConstructionService) ConstructionCombine(
 		return nil, wrapError(errInternalError, err)
 	}
 
+	wrappedSignedTx := signedTransactionWrapper{SignedTransaction: string(signedTxJSON), Currency: unsignedTx.Currency}
```
patrick-ogrady, line 195: Only possible nit could be that we are doing double marshaling here but 🤷 .

There is a small compatibility issue with a change like this where any inflight txs they have will no longer be "broadcastable" on this version.

patrick-ogrady: Something we'll want to call out

service/service_construction.go
```diff
@@ -272,23 +283,42 @@ func (s ConstructionService) ConstructionParse(
 	if !ok {
 		return nil, wrapError(errInvalidInput, fmt.Errorf("%s is not a valid address", tx.To))
 	}
+	var opMethod string
+	var value *big.Int
+	// Erc20 transfer
+	if len(tx.Data) != 0 {
+		if !mapper.EqualFoldContains(s.config.TokenWhiteList, tx.To) {
```
patrick-ogrady, line 290: This may be a bit of a dangerous check in case they modify this list with inflight txs. I think it is ok to not do this check in offline mode.

service/service_construction.go
```diff
@@ -397,10 +409,15 @@ func (s ConstructionService) ConstructionPayloads(
...
 	gasLimit := metadata.GasLimit
 	chainID := s.config.ChainID
-	transferData := []byte{}
 
 	fromOp, _ := matches[0].First()
 	fromAddress := fromOp.Account.Address
+	fromCurrency := fromOp.Amount.Currency
+	toCurrency := toOp.Amount.Currency
+
+	if types.Hash(fromCurrency) != types.Hash(toCurrency) {
```
patrick-ogrady, line 418: This will already be enforced by `MatchOperations`.

service/service_construction.go
```diff
@@ -412,14 +429,35 @@ func (s ConstructionService) ConstructionPayloads(
...
+			nonce,
+			ethcommon.HexToAddress(checkTo),
+			amount,
+			gasLimit,
+			gasPrice,
+			transferData,
+		)
+	} else {
+		contract, ok := fromCurrency.Metadata[mapper.ContractAddressMetadata]
+
```
patrick-ogrady, line 445: nit: extra \n

service/service_construction.go
```diff
@@ -412,14 +429,35 @@ func (s ConstructionService) ConstructionPayloads(
...
+		tx = ethtypes.NewTransaction(
+			nonce,
+			ethcommon.HexToAddress(checkTo),
+			amount,
+			gasLimit,
+			gasPrice,
+			transferData,
+		)
+	} else {
+		contract, ok := fromCurrency.Metadata[mapper.ContractAddressMetadata]
```
patrick-ogrady, line 444: `fromCurrency.Metadata[mapper.ContractAddressMetadata].(string)` so you don't need to do `.(string)` on 454. (this returns `!ok` if either the key doesn't exist or it is not a string)

service/service_construction.go
```diff
@@ -498,6 +520,13 @@ func (s ConstructionService) ConstructionPreprocess(
 	toOp, amount := matches[1].First()
 	toAddress := toOp.Account.Address
 
+	fromCurrency := fromOp.Amount.Currency
+	toCurrency := toOp.Amount.Currency
+
+	if types.Hash(fromCurrency) != types.Hash(toCurrency) {
```
patrick-ogrady, line 526: See above about `MatchOperations`

service/service_construction.go
```diff
@@ -588,3 +619,176 @@ func (s ConstructionService) ConstructionSubmit(
...
+	if types.Hash(firstCurrency) == types.Hash(mapper.AvaxCurrency) {
+		return s.createOperationDescriptionNative(), nil
+	}
+	// Not Native Avax, we require contractInfo in metadata
+	if firstCurrency.Metadata[mapper.ContractAddressMetadata] == nil ||
+		secondCurrency.Metadata[mapper.ContractAddressMetadata] == nil {
+		return nil, fmt.Errorf("non-native currency must have contractAddress in metadata")
+	}
+
+	contractAddress := firstCurrency.Metadata[mapper.ContractAddressMetadata]
```
patrick-ogrady, line 650: nit: make sure first/second match before using first

service/service_construction.go
patrick-ogrady: Good catch

service/service_construction.go
```diff
@@ -588,3 +619,176 @@ func (s ConstructionService) ConstructionSubmit(
...
+
+	send := parser.OperationDescription{
+		Type: mapper.OpErc20Transfer,
+		Account: &parser.AccountDescription{
+			Exists: true,
+		},
+		Amount: &parser.AmountDescription{
+			Exists:   true,
+			Sign:     parser.NegativeAmountSign,
+			Currency: mapper.Erc20Currency(currencyInfo.Symbol, currencyInfo.Decimals, contractAddress),
```
patrick-ogrady, line 697: nit: init currency once outside these vars

service/service_construction.go
```diff
@@ -588,3 +619,176 @@ func (s ConstructionService) ConstructionSubmit(
...
+	if err != nil {
+		return 0, err
+	}
+	return gasLimit, nil
+}
+
+func (s ConstructionService) getErc20TransferGasLimit(ctx context.Context, toAddress string,
+	fromAddress string, value *big.Int, currency *types.Currency) (uint64, error) {
+	contract, ok := currency.Metadata[mapper.ContractAddressMetadata]
+
```
patrick-ogrady, line 739: nit: extra \n

service/service_construction.go
```diff
@@ -588,3 +619,176 @@ func (s ConstructionService) ConstructionSubmit(
...
+	var data []byte
+	data = append(data, methodID...)
+	data = append(data, paddedAddress...)
+	data = append(data, paddedAmount...)
+	return data
+}
+
+func parseErc20TransferData(data []byte) (*ethcommon.Address, *big.Int, error) {
+	genericTransferBytesLength := 69
```
patrick-ogrady, line 773: make a const outside this func

service/service_construction.go
```diff
@@ -588,3 +619,176 @@ func (s ConstructionService) ConstructionSubmit(
...
+		return 0, err
+	}
+	return gasLimit, nil
+}
+
+func generateErc20TransferData(toAddress string, value *big.Int) []byte {
+	to := common.HexToAddress(toAddress)
+	methodID := getTransferMethodID()
+
+	requiredPaddingBytes := 32
```
patrick-ogrady, line 761: make a const outside this func

service/service_construction.go
```diff
@@ -588,3 +619,176 @@ func (s ConstructionService) ConstructionSubmit(
...
+	}
+
+	address := ethcommon.BytesToAddress(data[5:36])
+	amount := new(big.Int)
+	amount.SetBytes(data[37:])
+	return &address, amount, nil
+}
+
+func getTransferMethodID() []byte {
+	transferFnSignature := []byte("transfer(address,uint256)") // do not include spaces in the string
```
patrick-ogrady, line 789: make string a const outside this func

service/service_construction.go
learyce: Yeah, figured it was easier to just double marshal than worry about any of the eth transaction type.  Happy to change if you feel strongly

service/service_construction.go
learyce: We check the hash of the currencies to match above, which checks that the metadata matches as well

service/service_construction.go
patrick-ogrady: Nah this is fine.

client/client.go
```diff
@@ -50,23 +52,26 @@ type client struct {
...
 
-	contract, err := NewContractClient(endpoint)
+	contract, err := NewContractClient(endpointURL)
 	if err != nil {
 		return nil, err
 	}
 
 	return client{
-		Client:         info.NewClient(endpoint),
+		Client:         info.NewClient(endpointURL),
```
dhrubabasu, line 74: `endpoint` is correct here. The info client isn't an eth client.

cmd/server/config.go
```diff
@@ -91,3 +93,20 @@ func (c *config) Validate() error {
 
 	return nil
 }
+
+func (c *config) ValidateWhitelistOnlyValidErc20s(cli client.Client) error {
+	if len(c.TokenWhiteList) == 0 {
+		return nil
+	}
```
dhrubabasu, line 100: This check can be deleted

cmd/server/config.go
```diff
@@ -91,3 +93,20 @@ func (c *config) Validate() error {
...
+		ethAddress := ethcommon.HexToAddress(token)
+		contractInfo, err := cli.ContractInfo(ethAddress, true)
+		if err != nil {
+			return err
+		}
+		if contractInfo.Decimals == client.UnknownERC20Decimals && contractInfo.Symbol == client.UnknownERC20Symbol {
+			return errInvalidErc20Address
+		}
+	}
+	return nil
```
dhrubabasu, line 111:
```suggestion

	return nil
```

service/types.go
```diff
@@ -293,3 +303,33 @@ func (m *accountMetadata) UnmarshalJSON(data []byte) error {
 func has0xPrefix(str string) bool {
 	return len(str) >= 2 && str[0] == '0' && (str[1] == 'x' || str[1] == 'X')
 }
+
+type signedTransactionWrapper struct {
+	SignedTransaction string          `json:"signed_tx"`
+	Currency          *types.Currency `json:"currency,omitempty"`
+}
+
+type signedTransactionWrapperWire struct {
```
dhrubabasu, line 312: Why do we need this? It's identical to `signedTransactionWrapper`?

mapper/amount.go
```diff
@@ -41,15 +47,21 @@ func Erc20Amount(
...
-			Metadata: metadata,
-		},
+		Value:    decimalValue.String(),
+		Currency: currency,
+	}
+}
+
+func Erc20Currency(symbol string, decimals int32, contractAddress string) *types.Currency {
+	metadata := make(map[string]interface{})
+	metadata[ContractAddressMetadata] = contractAddress
```
dhrubabasu, line 60: You can instantiate maps inline:
```golang
map[string]interface{}{
	ContractAddressMetadata: contractAddress,
}
```

dhrubabasu: should also be moved into the return statement directly

dhrubabasu (changes requested):
left some nits

service/service_construction.go
```diff
@@ -588,3 +605,174 @@ func (s ConstructionService) ConstructionSubmit(
...
+) ([]*parser.OperationDescription, error) {
+	if len(operations) != 2 { //nolint:gomnd
+		return nil, fmt.Errorf("invalid number of operations")
+	}
+
+	firstCurrency := operations[0].Amount.Currency
+	secondCurrency := operations[1].Amount.Currency
+
+	if firstCurrency == nil || secondCurrency == nil {
+		return nil, fmt.Errorf("invalid currency on opeartion")
```
racbc, line 620: Typo in the error string.

service/service_construction.go
```diff
@@ -588,3 +605,174 @@ func (s ConstructionService) ConstructionSubmit(
...
+	paddedAmount := common.LeftPadBytes(value.Bytes(), requiredPaddingBytes)
+
+	var data []byte
+	data = append(data, methodID...)
+	data = append(data, paddedAddress...)
+	data = append(data, paddedAmount...)
+	return data
+}
+
+func parseErc20TransferData(data []byte) (*ethcommon.Address, *big.Int, error) {
```
learyce, line 757: Here

service/service_construction.go
```diff
@@ -430,6 +459,7 @@ func (s ConstructionService) ConstructionPayloads(
 		GasPrice: gasPrice,
```
patrick-ogrady, line 459: `to` should be `sendToAddress`

mapper/amount.go
learyce: Not sure about Golang, but it's usually best to declare complex objects then return.  If an error happens in the return statement, the call stack can get messedup

service/types.go
learyce: Following the convention above

service/service_construction.go
racbc, line 637: Comment is resolved but the typo is still there. Should be 'operation' instead of 'opeartion'.

service/service_construction.go
```diff
@@ -140,8 +145,13 @@ func (s ConstructionService) ConstructionHash(
 		return nil, wrapError(errInvalidInput, "signed transaction value is not provided")
 	}
 
+	wrappedTx := new(signedTransactionWrapper)
+	if err := json.Unmarshal([]byte(req.SignedTransaction), wrappedTx); err != nil {
```
dhrubabasu, line 149:
```suggestion
	var wrappedTx signedTransactionWrapper
	if err := json.Unmarshal([]byte(req.SignedTransaction), &wrappedTx); err != nil {
```

dhrubabasu: There's other places in the PR where we use `new` when we should be using `var` like this.

service/service_construction.go
```diff
@@ -588,3 +621,176 @@ func (s ConstructionService) ConstructionSubmit(
 		},
 	}, nil
 }
+
+func (s ConstructionService) CreateOperationDescription(
+	operations []*types.Operation,
+) ([]*parser.OperationDescription, error) {
+	if len(operations) != 2 { //nolint:gomnd
+		return nil, fmt.Errorf("invalid number of operations")
```
dhrubabasu, line 629: Error should be a variable

service/service_construction.go
```diff
@@ -588,3 +621,176 @@ func (s ConstructionService) ConstructionSubmit(
...
+	}
+
+	firstCurrency := operations[0].Amount.Currency
+	secondCurrency := operations[1].Amount.Currency
+
+	if firstCurrency == nil || secondCurrency == nil {
+		return nil, fmt.Errorf("invalid currency on opeartion")
```
dhrubabasu, line 636: Error should be a variable, also operation is mispelled

dhrubabasu: There's other errors that aren't variables below

service/service_construction.go
dhrubabasu: This wasn't resolved

service/service_construction.go
dhrubabasu: This wasn't resolved

dhrubabasu (changes requested):
Left some nits

service/service_construction.go
learyce: What's the difference?

dhrubabasu: It's purely stylistic, we use `var` in the codebase so we should be consistent

service/service_construction.go
learyce: It's an invalid input error, it's wrapped in the function calling this one (same with the others).  Doesn't seem like we need a unique error type for every way the input could be invalid

learyce: Can't find my previous comment, but the caller is wrapping all the errors you mentioned within an invalidInputErr.  We don't want unique errors defined for each way that the input can be incorrect

service/service_construction.go
learyce: updated

service/service_construction.go
```diff
@@ -782,9 +782,9 @@ func parseErc20TransferData(data []byte) (*ethcommon.Address, *big.Int, error) {
 	}
 
 	address := ethcommon.BytesToAddress(data[5:36])
-	amount := new(big.Int)
+	var amount big.Int
 	amount.SetBytes(data[37:])
```
patrick-ogrady, line 786: you should do `amount := new(big.Int).SetBytes(data[37:])`

patrick-ogrady: you shouldn't init a `big.Int` not as a pointer

PR #66, opened 2022-01-13
cleanup `client` package

dhrubabasu:


client/types.go
```diff
@@ -21,7 +21,7 @@ type infoPeersResponse struct {
 
 type ContractInfo struct {
 	Symbol   string `json:"symbol"`
-	Decimals uint8  `json:"decimals"`
+	Decimals int32  `json:"decimals"`
```
patrick-ogrady, line 24: Why not just do `uint` or `int` here? by default it is already 32.

dhrubabasu: removed it altogether and am just using `types.Currency`: 0d901a2

client/contract.go
```diff
@@ -31,35 +32,37 @@ func NewContractClient(endpointURL string) (*ContractClient, error) {
...
 			symbol = UnknownERC20Symbol
 			decimals = UnknownERC20Decimals
 		} else {
 			symbol = UnknownERC721Symbol
 			decimals = UnknownERC721Decimals
 		}
 	}
-	contractInfo := &ContractInfo{Symbol: symbol, Decimals: decimals}
+
+	currency := &types.Currency{
```
patrick-ogrady, line 60: I don't like the partial population of this. We should use the existing code (with a different var name)

dhrubabasu: resolved in 1c42325

dhrubabasu:
ready for review now

client/client.go
```diff
@@ -60,11 +58,6 @@ func NewClient(endpoint string) (Client, error) {
 		return nil, err
 	}
 
-	evmlogs, err := NewEvmLogsClient(endpointURL)
-	if err != nil {
-		return nil, err
-	}
-
 	contract, err := NewContractClient(endpointURL)
```
patrick-ogrady, line 61: Is there a reason we can't pass the client in here?

dhrubabasu: fixed in 14bdc3d

patrick-ogrady (approved):

PR #77, opened 2022-02-02
cleanup types

dhrubabasu:


mapper/transaction.go
```diff
@@ -510,8 +509,8 @@ func parseErc721Txs(transferLog ethtypes.Log, opsLen int64) []*types.Operation {
 	addressTo := transferLog.Topics[2]
 	erc721Index := transferLog.Topics[3] // Erc721 4th topic is the index.  Data is empty
 	metadata := map[string]interface{}{
-		ContractAddressMetadata:  contractAddress.String(),
-		IndexTransferredMetadata: erc721Index.String(),
+		clientTypes.ContractAddressMetadata: contractAddress.String(),
+		IndexTransferredMetadata:            erc721Index.String(),
```
dhrubabasu, line 513: This is where we manually set `indexTransferred` for ERC-721s

mapper/transaction.go
```diff
@@ -506,8 +506,8 @@ func parseErc721Txs(transferLog ethtypes.Log, opsLen int64) []*types.Operation {
 	addressTo := transferLog.Topics[2]
 	erc721Index := transferLog.Topics[3] // Erc721 4th topic is the index.  Data is empty
 	metadata := map[string]interface{}{
-		ContractAddressMetadata:  contractAddress.String(),
-		IndexTransferredMetadata: erc721Index.String(),
+		"contractAddress":  contractAddress.String(),
+		"indexTransferred": erc721Index.String(),
```
dhrubabasu, line 510: should we change `contractAddress` -> `contract_address` and `indexTransferred` to `index_transferred`? All the other metadata maps use snake case rather than camel case.

mapper/types.go
```diff
@@ -13,11 +13,6 @@ const (
 	FujiChainID = 43113
 	FujiAssetID = "U8iRqJoiJm8xZHAacmvYyZVwqQx6uDNtQeP3CQ6fcgQk3JqnK"
 
-	TokenTypeMetadata        = "tokenType"
```
patrick-ogrady, line 16: Why are we removing constants?

dhrubabasu: `TokenTypeMetadata` and `TokenSymbol` are unused. For metadata maps, we haven't used constants elsewhere in the codebase as it sometimes creates circular dependencies.

I can put them back if you'd like but seemed odd to make an exception for just these two

PR #84, opened 2022-02-07
[CPT-221] Use `debug_traceBlock` for `/block`

dhrubabasu:


client/eth.go
```diff
@@ -50,15 +47,35 @@ func (c *EthClient) TxPoolContent(ctx context.Context) (*TxPoolContent, error) {
 	return content, err
 }
 
-// TraceTransaction returns a transaction trace
```
learyce, line 53: Why did you kill all the comments?

client/eth.go
```diff
@@ -50,15 +47,35 @@ func (c *EthClient) TxPoolContent(ctx context.Context) (*TxPoolContent, error) {
...
 func (c *EthClient) TraceTransaction(ctx context.Context, hash string) (*Call, []*FlatCall, error) {
-	result := &Call{}
+	var result *Call
```
learyce, line 51: What's the difference between these two ways to init in Golang?

service/service_block.go
```diff
@@ -140,7 +140,12 @@ func (s *BlockService) BlockTransaction(
...
 	}
 
-	transaction, terr := s.fetchTransaction(ctx, tx, header)
+	trace, flattened, err := s.client.TraceTransaction(ctx, tx.Hash().String())
+	if err != nil {
+		return nil, wrapError(errClientError, err)
+	}
+
+	transaction, terr := s.fetchTransaction(ctx, tx, header, trace, flattened)
 	if err != nil {
```
learyce, line 149: Should this be terr ?

service/service_block.go
```diff
@@ -156,11 +161,14 @@ func (s *BlockService) fetchTransactions(
 ) ([]*types.Transaction, *types.Error) {
 	transactions := []*types.Transaction{}
 
-	for _, tx := range block.Transactions() {
-		transaction, err := s.fetchTransaction(ctx, tx, block.Header())
+	trace, flattened, err := s.client.TraceBlockByHash(ctx, block.Hash().String())
+
+	for i, tx := range block.Transactions() {
+		transaction, terr := s.fetchTransaction(ctx, tx, block.Header(), trace[i], flattened[i])
 		if err != nil {
```
learyce, line 168: Should this be terr ?

learyce, line 168: Should this be terr ?

client/eth.go
dhrubabasu: These comments are pretty useless imo, reverted in 5d40801 (will address in another PR)

service/service_block.go
dhrubabasu: Yes, fixed in 6f5d163

dhrubabasu: Yes, fixed in 6f5d163

service/service_block.go
dhrubabasu: Yes, fixed in 6f5d163

learyce (dismissed):

client/eth.go
patrick-ogrady: This won't work @dhrubabasu. This just creates an unallocated pointer (is `nil` not `&Call{}`)

dhrubabasu: Addressed in a9525f4

patrick-ogrady: should do `var result Call`

patrick-ogrady (approved):
LGTM

PR #89, opened 2022-02-08
cleanup transaction mapper

dhrubabasu:


learyce:
Looks like an improvement, but can we add a few tests to the mapper file to make sure we don't accidentally cause a regression

dhrubabasu:
> Looks like real improvement, but can we add a few tests to the mapper file to make sure we don't accidentally cause a regression

done!

patrick-ogrady (approved):
LGTM. Thanks for adding tests.

mapper/transaction.go
```diff
@@ -439,133 +442,118 @@ func traceOps(trace []*clientTypes.FlatCall, startIndex int) []*types.Operation
 }
 
 func erc20Ops(transferLog *ethtypes.Log, currency *clientTypes.ContractCurrency, opsLen int64) []*types.Operation {
-	ops := []*types.Operation{}
-
 	contractAddress := transferLog.Address
-	addressFrom := transferLog.Topics[1]
-	addressTo := transferLog.Topics[2]
+	fromAddress := common.BytesToAddress(transferLog.Topics[1].Bytes())
```
patrick-ogrady, line 446: Very nuanced behavior on this guy: https://github.com/ethereum/go-ethereum/blob/830231c1c40f45f32cbf4ab69954b87a79313f44/common/types.go#L203-L204

learyce (approved):

PR #94, opened 2022-02-09
[CPT-247] add support for legacy signed tx format

learyce:


service/types.go
```diff
@@ -305,3 +307,37 @@ type signedTransactionWrapper struct {
...
+
+func (t *signedTransactionWrapper) UnmarshalJSON(data []byte) error {
+	var tw signedTransactionWrapperWire
+	if err := json.Unmarshal(data, &tw); err != nil {
+		return err
+	}
+
+	if tw.SignedTransaction == nil {
+		//Check if unmarshal is empty because of legacy format
+		var signedTx ethtypes.Transaction
```
learyce, line 332: Do we need to check the values of this unmarshalled object?

service/types.go
```diff
@@ -305,3 +307,37 @@ type signedTransactionWrapper struct {
...
+		if err := signedTx.UnmarshalJSON(data); err == nil {
```
dhrubabasu, line 333: This error should be returned if not nil

dhrubabasu (changes requested):
LGTM aside from lint errors

service/types.go
learyce: I don't think we want to return an error here.  We return an error if the original unmarshalling fails.  This is an optional check to see if we're in the old format.  If this errors, we should just proceed as if nothing happened (hence the if err == nil, assume legacy, else, pretend we never checked).

dhrubabasu: We want to return an error here, it is very confusing UX wise why it fails otherwise:


```
// PR as-is
Error:      	Expected nil, but got: &types.Error{Code:6, Message:"Invalid input", Description:(*string)(nil), Retriable:false, Details:map[string]interface {}{"error":"unexpected end of JSON input"}}

// returning the error
Error:      	Expected nil, but got: &types.Error{Code:6, Message:"Invalid input", Description:(*string)(nil), Retriable:false, Details:map[string]interface {}{"error":"missing required field 'nonce' in transaction"}}
```

We should also add a test for invalid input legacy transactions to verify the error response.

service/types.go
```diff
@@ -305,3 +307,38 @@ type signedTransactionWrapper struct {
 	SignedTransaction []byte          `json:"signed_tx"`
 	Currency          *types.Currency `json:"currency,omitempty"`
 }
+
+type signedTransactionWrapperWire struct {
+	SignedTransaction []byte          `json:"signed_tx"`
+	Currency          *types.Currency `json:"currency,omitempty"`
+}
+
+func (t *signedTransactionWrapper) MarshalJSON() ([]byte, error) {
```
patrick-ogrady, line 316: You don't need to override `MarshalJSON` because it doesn't change default behavior.

dhrubabasu (approved):

PR #99, opened 2022-03-16
remove `ContractCurrency` type

dhrubabasu:
We maintain an invariant that `contractAddress` is always populated in ERC-20 metadata making this type unnecessary.

ERC-721s don't use currencies since they are transferred by index.

### DIFF
--- a/client/contract.go
+++ b/client/contract.go
@@ -3,7 +3,6 @@ package client
 import (
 	"github.com/ava-labs/avalanchego/cache"
 	"github.com/ava-labs/coreth/ethclient"
-	"github.com/coinbase/rosetta-sdk-go/types"
 	"github.com/ethereum/go-ethereum/common"
 )
 
@@ -25,15 +24,23 @@ func NewContractClient(c ethclient.Client) *ContractClient {
 	}
 }
 
-// GetContractCurrency returns the currency for a specific address
-func (c *ContractClient) GetContractCurrency(addr common.Address, erc20 bool) (*types.Currency, error) {
+// GetContractInfo returns the symbol and decimals for [addr].
+func (c *ContractClient) GetContractInfo(addr common.Address, erc20 bool) (string, uint8, error) {
+	// We don't define another struct because this is never used outside of this
+	// function.
+	type ContractInfo struct {
+		Symbol   string
+		Decimals uint8
+	}
+
 	if currency, cached := c.cache.Get(addr); cached {
-		return currency.(*types.Currency), nil
+		cast := currency.(*ContractInfo)
+		return cast.Symbol, cast.Decimals, nil
 	}
 
 	token, err := NewContractInfoToken(addr, c.ethClient)
 	if err != nil {
-		return nil, err
+		return "", 0, err
 	}
 
 	symbol, symbolErr := token.Symbol(nil)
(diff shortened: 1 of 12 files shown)


client/contract.go
```diff
@@ -25,9 +26,9 @@ func NewContractClient(c ethclient.Client) *ContractClient {
 }
 
 // GetContractCurrency returns the currency for a specific address
-func (c *ContractClient) GetContractCurrency(addr common.Address, erc20 bool) (*ContractCurrency, error) {
+func (c *ContractClient) GetContractCurrency(addr common.Address, erc20 bool) (*types.Currency, error) {
```
patrick-ogrady, line 29: I have thus far avoid brining any knowledge of `Rosetta` into `client`. This breaks that abstraction.

patrick-ogrady: I think we should instead add a function in `mapper` that wraps this and returns the `Currency`. The simplifications in `mapper` are on-track.

patrick-ogrady: Still reconciling this opinion with the need to keep an extra type around.

dhrubabasu: Good point, I've provided an alternative way in 656d0cd that respects the abstraction

mapper/transaction.go
```diff
@@ -88,28 +88,28 @@ func Transaction(
...
 				return nil, err
 			}
 
-			if currency.Symbol == clientTypes.UnknownERC20Symbol && !includeUnknownTokens {
+			if symbol == clientTypes.UnknownERC20Symbol && !includeUnknownTokens {
 				continue
 			}
 
-			erc20Ops := erc20Ops(log, currency, int64(len(ops)))
+			erc20Ops := erc20Ops(log, ToCurrency(symbol, decimals, log.Address), int64(len(ops)))
```
dhrubabasu, line 112: was going back and forth on whether or not to wrap the `client.GetContractInfo` call in `ToCurrency` but I think the code is a lot more readable this way (especially since we only use `types.Currency` for erc20s.

client/contract.go
```diff
@@ -24,37 +24,42 @@ func NewContractClient(c ethclient.Client) *ContractClient {
...
-		return nil, err
+		return "", 0, err
 	}
 
-	symbol, symbolErr := token.Symbol(nil)
-	decimals, decimalErr := token.Decimals(nil)
-
-	// Any of these indicate a failure to get complete information from contract
-	if symbolErr != nil || decimalErr != nil || symbol == "" || decimals == 0 {
+	// [symbol] is set to "" if [token.Symbol] errors.
```
dhrubabasu, line 46: https://github.com/ava-labs/avalanche-rosetta/blob/master/client/contractInfoToken.go#L216-L228

client/contract.go
```diff
@@ -24,37 +24,42 @@ func NewContractClient(c ethclient.Client) *ContractClient {
...
 			symbol = UnknownERC721Symbol
-			decimals = UnknownERC721Decimals
 		}
 	}
 
-	currency := &ContractCurrency{
-		Symbol:   symbol,
-		Decimals: int32(decimals),
-	}
+	// [decimals] is set to 0 if [token.Decimals] errors.
```
dhrubabasu, line 56: https://github.com/ava-labs/avalanche-rosetta/blob/master/client/contractInfoToken.go#L185-L197

client/contract.go
```diff
@@ -24,37 +24,42 @@ func NewContractClient(c ethclient.Client) *ContractClient {
...
+	decimals, _ := token.Decimals(nil)
```
patrick-ogrady, line 57: There can also be tokens with no decimals though? In that case, you are silencing the errror.

mapper/transaction.go
patrick-ogrady: Yeah, I agree.

client/contract.go
dhrubabasu: The existing code only returns an error if `token` fails to initialize: https://github.com/ava-labs/avalanche-rosetta/blob/master/client/contract.go#L33-L36

if `token.Decimals()` ever errors, we default to 0: https://github.com/ava-labs/avalanche-rosetta/blob/master/client/types.go#L11-L15

Since it's already set to 0 on error (via `*new(uint8)`), we don't need to check this error: https://github.com/ava-labs/avalanche-rosetta/blob/master/client/contractInfoToken.go#L185-L197

patrick-ogrady (approved):

PR #101, opened 2022-03-18
cleanup erc20 methods

dhrubabasu:


service/service_construction.go
```diff
@@ -22,10 +23,8 @@ import (
 )
 
 const (
-	genericTransferBytesLength = 68
-	requiredPaddingBytes       = 32
```
learyce, line 26: Let's keep these as non-magicNumbers

learyce, line 26: We should keep these as non-magicNumbers

service/service_construction.go
```diff
@@ -628,142 +627,109 @@ func (s ConstructionService) CreateOperationDescription(
 		return nil, fmt.Errorf("invalid number of operations")
 	}
 
-	firstCurrency := operations[0].Amount.Currency
-	secondCurrency := operations[1].Amount.Currency
+	currency := operations[0].Amount.Currency
```
learyce, line 630: I'm not sure why we removed the firstCurrency/SecondCurrency to instead keep one broken out and reference the other one within the array

service/service_construction.go
```diff
@@ -628,142 +627,109 @@ func (s ConstructionService) CreateOperationDescription(
...
+		return s.createOperationDescription(currency, mapper.OpCall), nil
 	}
-	_, firstOk := firstCurrency.Metadata[mapper.ContractAddressMetadata].(string)
-	_, secondOk := secondCurrency.Metadata[mapper.ContractAddressMetadata].(string)
 
-	// Not Native Avax, we require contractInfo in metadata
-	if !firstOk || !secondOk {
-		return nil, fmt.Errorf("non-native currency must have contractAddress in metadata")
+	// ERC-20s must have contract address in metadata
+	if _, ok := currency.Metadata[mapper.ContractAddressMetadata].(string); !ok {
```
learyce, line 645: Agree with the chance here, but a comment about why we only need to check the first item would be helpful

service/service_construction.go
```diff
@@ -628,142 +627,109 @@ func (s ConstructionService) CreateOperationDescription(
...
-	nativeReceive := parser.OperationDescription{
-		Type: mapper.OpCall,
-		Account: &parser.AccountDescription{
-			Exists: true,
-		},
-		Amount: &parser.AmountDescription{
-			Exists:   true,
-			Sign:     parser.PositiveAmountSign,
-			Currency: mapper.AvaxCurrency,
+func (s ConstructionService) createOperationDescription(
```
learyce, line 652: Solid change

service/service_construction.go
```diff
@@ -628,142 +627,109 @@ func (s ConstructionService) CreateOperationDescription(
...
-
-	descriptions = append(descriptions, &send)
-	descriptions = append(descriptions, &receive)
-	return descriptions
 }
 
 func (s ConstructionService) getNativeTransferGasLimit(
-	ctx context.Context, toAddress string,
-	fromAddress string, value *big.Int,
+	ctx context.Context,
```
learyce, line 686: :+1:

service/service_construction.go
```diff
@@ -772,24 +738,27 @@ func generateErc20TransferData(toAddress string, value *big.Int) []byte {
...
+// Ref: https://goethereumbook.org/en/transfer-tokens/#forming-the-data-field
 func parseErc20TransferData(data []byte) (*ethcommon.Address, *big.Int, error) {
-	if len(data) != genericTransferBytesLength {
+	if len(data) != 68 {
 		return nil, nil, fmt.Errorf("incorrect length for data array")
 	}
-	methodID := getTransferMethodID()
-	if hexutil.Encode(data[:4]) != hexutil.Encode(methodID) {
+
+	methodBytes, addrBytes, amtBytes := data[:4], data[5:36], data[37:]
```
learyce, line 747: :smile:

service/service_construction.go
dhrubabasu: Where would you put the comment?

service/service_construction.go
dhrubabasu: After the utils.Equal call on 639, we only need to reference one of them since they're identical

service/service_construction.go
dhrubabasu: I think we can use the abi package to make this more robust - will try in a follow-up PR

learyce (approved):

PR #103, opened 2022-03-18
cleanup `CreateOperationDescription`

dhrubabasu:


service/service_construction.go
```diff
@@ -621,29 +622,26 @@ func (s ConstructionService) CreateOperationDescription(
...
+		return s.createOperationDescription(currency, mapper.OpCall), nil
 	}
-	_, firstOk := firstCurrency.Metadata[mapper.ContractAddressMetadata].(string)
-	_, secondOk := secondCurrency.Metadata[mapper.ContractAddressMetadata].(string)
 
-	// Not Native Avax, we require contractInfo in metadata
-	if !firstOk || !secondOk {
-		return nil, fmt.Errorf("non-native currency must have contractAddress in metadata")
+	// ERC-20s must have contract address in metadata
+	if _, ok := currency.Metadata[mapper.ContractAddressMetadata].(string); !ok {
```
patrick-ogrady, line 640: We should check both currencies

dhrubabasu: we verify they're equivalent in line 631 which includes the `Metadata` map:
```golang
	if !utils.Equal(currency, operations[1].Amount.Currency) {
```

patrick-ogrady: Ah fair.

patrick-ogrady (approved):

PR #115, opened 2022-05-21
use `coreth/rpc`

dhrubabasu:


client/client.go
```diff
@@ -45,17 +45,14 @@ type client struct {
...
-	eth, err := NewEthClient(endpoint)
-	if err != nil {
-		return nil, err
-	}
+	eth, err := NewEthClient(ctx, endpoint)
 
 	return client{
 		Client:         info.NewClient(endpoint),
 		EthClient:      eth,
 		ContractClient: NewContractClient(eth.Client),
```
patrick-ogrady, line 56: This will panic as `eth` could be `nil`

client/eth.go
```diff
@@ -18,52 +18,41 @@ var (
...
-	c, err := ethclient.Dial(endpointURL)
-	if err != nil {
-		return nil, err
-	}
+	c, err := rpc.DialContext(ctx, endpointURL)
 
 	return &EthClient{
-		Client: c,
-		rpc:    rpc.NewRPCRequester(endpoint),
+		Client: ethclient.NewClient(c),
```
patrick-ogrady, line 32: Could panic if `c` is nil...just return err

dhrubabasu: This wouldn't panic - reverting these changes so it's more obvious

patrick-ogrady (approved):
Tested on mainnet 👍

PR #123, opened 2022-07-19
P-Chain Support

rajranjan0608:


service/backend/pchain/backend.go
```diff
@@ -0,0 +1,110 @@
...
+	assetID ids.ID,
+	networkIdentifier *types.NetworkIdentifier,
+) *Backend {
+	return &Backend{
+		fac:               &crypto.FactorySECP256K1R{},
+		networkIdentifier: networkIdentifier,
+		pClient:           pClient,
+		getUTXOsPageSize:  1024,
+		codec:             blocks.Codec,
+		codecVersion:      0,
```
abi87, line 51: nit: maybe use `codecVersion: txs.Version,` rather than hardcoding 0, as we do in avalanchego codebase

service/backend/pchain/indexer/types.go
```diff
@@ -0,0 +1,36 @@
...
+	UTXOs         []*genesis.UTXO `json:"utxos"`
+}
+
+type ParsedGenesisBlock struct {
+	ParsedBlock
+	GenesisBlockData `json:"data"`
+}
+
+type Proposer struct {
+	ID           ids.ID     `json:"id"`
```
abi87, line 31: q: I see the proposer certificate is not considered here. Is it on purpose? Note that certificate are *not* displayed in every blocks (there won't be a proposer in case of blocks built past proposer window)

service/backend/pchain/indexer/parser.go
```diff
@@ -0,0 +1,318 @@
...
+	"github.com/ava-labs/avalanche-rosetta/mapper"
+	"github.com/ava-labs/avalanche-rosetta/service/backend/common"
+)
+
+var errParserUninitialized = errors.New("uninitialized parser")
+
+const genesisTimestamp = 1599696000
+
+type Parser interface {
+	Initialize(ctx context.Context) (*ParsedGenesisBlock, error)
```
abi87, line 31: q: It feels weird having an `Initialize` method returning the genesis. Why not splitting the initiazation from the retrieval of genesis block, which may be carried out with e.g.  `ParseBlockAtIndex`?

service/backend/pchain/block.go
```diff
@@ -0,0 +1,431 @@
...
+		}
+
+		transactions, err = b.parseTransactions(ctx, request.NetworkIdentifier, block.Txs)
+		if err != nil {
+			return nil, service.WrapError(service.ErrInternalError, err)
+		}
+	}
+
+	for i := range transactions {
+		transaction := transactions[i]
```
abi87, line 129: nit `	for _, transaction := range transactions {`
range returns both the index and the actual value

service/backend/pchain/block.go
```diff
@@ -0,0 +1,431 @@
...
+		}
+		dependencyTxIDs = append(dependencyTxIDs, inputTxsIds...)
+	}
+
+	dependencyTxChan := make(chan *pmapper.DependencyTx, len(dependencyTxIDs))
+	eg, ctx := errgroup.WithContext(ctx)
+
+	dependencyTxs := make(map[string]*pmapper.DependencyTx)
+	for i := range dependencyTxIDs {
+		txID := dependencyTxIDs[i]
```
abi87, line 188: same as above, you can shortcut this to
`	for _, txID := range dependencyTxIDs {`

abi87, line 188: An alternative is
```
for _, txID := range dependencyTxIDs {
	txID := txID
```
the second line needed to avoid capturing from the goroutine

mapper/pchain/tx_parser.go
```diff
@@ -0,0 +1,740 @@
...
+	return txIds, nil
+}
+
+func getUniqueTxIds(ins []*avax.TransferableInput) []ids.ID {
+	txnIDs := make(map[string]ids.ID)
+	for _, in := range ins {
+		txnIDs[in.UTXOID.TxID.String()] = in.UTXOID.TxID
+	}
+
+	uniqueTxnIDs := []ids.ID{}
```
abi87, line 629: nit: `uniqueTxnIDs := make([]ids.ID, 0, len(txnIDs))`

mapper/pchain/tx_parser.go
```diff
@@ -0,0 +1,740 @@
...
+				return nil, err
+			}
+		}
+	}
+
+	return addresses, nil
+}
+
+func GetDependencyTxIDs(tx txs.UnsignedTx) ([]ids.ID, error) {
+	var txIds []ids.ID
```
abi87, line 598: the new Blueberry transactions are missing here. I do not understand why

service/backend/pchain/block.go
```diff
@@ -0,0 +1,431 @@
...
+		return err
+	}
+
+	var tx txs.Tx
+	_, err = b.codec.Unmarshal(txBytes, &tx)
+	if err != nil {
+		return err
+	}
+
+	err = common.InitializeTx(0, blocks.Codec, &tx)
```
abi87, line 229: nit avoid hardcoding `0`, use `txs.Version`.

service/backend/pchain/block.go
```diff
@@ -0,0 +1,431 @@
...
+	ctx context.Context,
+	networkIdentifier *types.NetworkIdentifier,
+) (*indexer.ParsedGenesisBlock, []*types.Transaction, error) {
+	genesisBlock, err := b.getGenesisBlock(ctx)
+	if err != nil {
+		return nil, nil, err
+	}
+
+	genesisTxs := []*txs.Tx{}
+	genesisTxs = append(genesisTxs, genesisBlock.Txs...)
```
abi87, line 370: nit: `genesisTxs := genesisBlock.Txs`.

service/backend/pchain/block.go
```diff
@@ -0,0 +1,431 @@
...
+		return nil, nil, err
+	}
+	genesisTxs = append(genesisTxs, allocationTx)
+
+	parser, err := b.newTxParser(ctx, networkIdentifier, nil)
+	if err != nil {
+		return nil, nil, err
+	}
+
+	transactions := []*types.Transaction{}
```
abi87, line 383: nit: preallocate capacity for efficiency `transactions := make([]*types.Transaction, 0, len(genesisTxs))`

cmd/server/main.go
```diff
@@ -9,13 +9,18 @@ import (
 	"math/big"
```
abi87, line 9: Not really in the scope of this PR, but linter reports the following warning:
`cmd/server/main.go:7:2: `io/ioutil` is in the denylist: io/ioutil is deprecated. Use package io or os instead. (depguard)`

service/backend/pchain/construction.go
```diff
@@ -0,0 +1,333 @@
...
+	case *txs.CreateChainTx:
+		return utx.Ins, nil
+	case *txs.CreateSubnetTx:
+		return utx.Ins, nil
+	case *txs.ImportTx:
+		return utx.ImportedInputs, nil
+	case *txs.ExportTx:
+		return utx.Ins, nil
+	default:
+		return nil, errUnknownTxType
```
abi87, line 281: nit: I'd list explictly all of the transactions here and maybe find another error than `errUnknownTxType` for them

service/backend/pchain/indexer/parser.go
```diff
@@ -0,0 +1,318 @@
...
+	errs := wrappers.Errs{}
+
+	proposer, bytes, err := getProposerFromBytes(proposerBytes)
+	if err != nil {
+		return nil, fmt.Errorf("fetching proposer from block bytes errored with %w", err)
+	}
+
+	if p.genesisTimestamp.IsZero() {
+		return nil, errParserUninitialized
+	}
```
abi87, line 208: nit: move it before pulling proposer up, for an early return in case parser it's not duly initialized

service/backend/pchain/indexer/parser.go
```diff
@@ -0,0 +1,318 @@
...
+	} else {
+		*blockTimestamp = time.Unix(genesisTimestamp, 0)
+	}
+
+	var blk blocks.Block
+	ver, err := p.codec.Unmarshal(bytes, &blk)
+	if err != nil {
+		return nil, fmt.Errorf("unmarshaling block bytes errored with %w", err)
+	}
+	blkID := ids.ID(hashing.ComputeHash256Array(bytes))
```
abi87, line 224: consider `blkID := blk.ID()` here. In general it'd be best to pull IDs out of blocks rather than recalcutiong them here. This should simplify thing should ID implementation ever change

service/backend/pchain/indexer/parser.go
```diff
@@ -0,0 +1,318 @@
...
+	case *blocks.BlueberryStandardBlock:
+		var blockTxs []*txs.Tx
+
+		for i, tx := range castBlk.Transactions {
+			errs.Add(common.InitializeTx(ver, p.codec, tx))
+			blockTxs = append(blockTxs, castBlk.Transactions[i])
+		}
+
+		parsedBlock.ParentID = castBlk.PrntID
+		parsedBlock.Txs = blockTxs
```
abi87, line 268: nit:
```
	case *blocks.ApricotStandardBlock:
		for _, tx := range castBlk.Transactions {
			errs.Add(common.InitializeTx(ver, p.codec, tx))
		}
		parsedBlock.ParentID = castBlk.PrntID
		parsedBlock.Txs = castBlk.Transactions

	case *blocks.BlueberryStandardBlock:
		for _, tx := range castBlk.Transactions {
			errs.Add(common.InitializeTx(ver, p.codec, tx))
		}
		parsedBlock.ParentID = castBlk.PrntID
		parsedBlock.Txs = castBlk.Transactions
```

service/backend/pchain/indexer/parser.go
```diff
@@ -0,0 +1,318 @@
...
+		parsedBlock.ParentID = castBlk.PrntID
+		parsedBlock.Txs = []*txs.Tx{}
+	case *blocks.ApricotCommitBlock:
+		parsedBlock.ParentID = castBlk.PrntID
+		parsedBlock.Txs = []*txs.Tx{}
+	default:
+		errs.Add(fmt.Errorf("no handler exists for block type %T", castBlk))
+	}
+
+	// If the block has an advance time tx, use its timestamp as the block timestamp
```
abi87, line 285: Actually all blueberry blocks have their timestamp explicitly serialized. Moreover in Apricot blocks, only ApricotProposal ones may contain advanceTimeTx (which is then forbidden in Blueberry).

service/backend/pchain/indexer/parser.go
```diff
@@ -0,0 +1,318 @@
...
+	case *blocks.ApricotProposalBlock:
+		errs.Add(common.InitializeTx(ver, p.codec, castBlk.Tx))
+
+		parsedBlock.ParentID = castBlk.PrntID
+		parsedBlock.Txs = []*txs.Tx{castBlk.Tx}
+	case *blocks.BlueberryProposalBlock:
+		errs.Add(common.InitializeTx(ver, p.codec, castBlk.Tx))
+
+		parsedBlock.ParentID = castBlk.PrntID
+		parsedBlock.Txs = []*txs.Tx{castBlk.Tx}
```
abi87, line 243: insert: `parsedBlock.Timestamp = castBlk.Timestamp().Unix()`

service/backend/pchain/indexer/parser.go
abi87, line 268: nit:
```
	case *blocks.ApricotStandardBlock:
		for _, tx := range castBlk.Transactions {
			errs.Add(common.InitializeTx(ver, p.codec, tx))
		}
		parsedBlock.ParentID = castBlk.PrntID
		parsedBlock.Txs = castBlk.Transactions

	case *blocks.BlueberryStandardBlock:
		for _, tx := range castBlk.Transactions {
			errs.Add(common.InitializeTx(ver, p.codec, tx))
		}
		parsedBlock.ParentID = castBlk.PrntID
		parsedBlock.Txs = castBlk.Transactions
		parsedBlock.Timestamp = castBlk.Timestamp().Unix()
```

service/backend/pchain/indexer/parser.go
```diff
@@ -0,0 +1,318 @@
...
+		}
+
+		parsedBlock.ParentID = castBlk.PrntID
+		parsedBlock.Txs = blockTxs
+	case *blocks.ApricotAbortBlock:
+		parsedBlock.ParentID = castBlk.PrntID
+		parsedBlock.Txs = []*txs.Tx{}
+	case *blocks.BlueberryAbortBlock:
+		parsedBlock.ParentID = castBlk.PrntID
+		parsedBlock.Txs = []*txs.Tx{}
```
abi87, line 274: insert: `parsedBlock.Timestamp = castBlk.Timestamp().Unix()`

service/backend/pchain/indexer/parser.go
```diff
@@ -0,0 +1,318 @@
...
+	case *blocks.BlueberryCommitBlock:
+		parsedBlock.ParentID = castBlk.PrntID
+		parsedBlock.Txs = []*txs.Tx{}
```
abi87, line 277: insert: `parsedBlock.Timestamp = castBlk.Timestamp().Unix()`

service/backend/common/parser.go
```diff
@@ -0,0 +1,22 @@
+package common
+
+import (
+	"github.com/ava-labs/avalanchego/codec"
+	"github.com/ava-labs/avalanchego/utils/wrappers"
+	"github.com/ava-labs/avalanchego/vms/platformvm/txs"
+)
+
+// initializes tx to have tx identifier generated
+func InitializeTx(version uint16, c codec.Manager, tx *txs.Tx) error {
```
abi87, line 10: nit: seems only used in pchain package. Why is it in common?

service/backend/cchainatomictx/backend.go
```diff
@@ -0,0 +1,71 @@
...
+	avaxAssetID      ids.ID
+}
+
+func NewBackend(cClient client.Client, avaxAssetID ids.ID) *Backend {
+	return &Backend{
+		fac:              &crypto.FactorySECP256K1R{},
+		cClient:          cClient,
+		getUTXOsPageSize: 1024,
+		codec:            evm.Codec,
+		codecVersion:     0,
```
abi87, line 36: this should be filled with evm.codecVersion which is not exported. We should probably ask devs to export it, similarly to what it is done with pchain

abi87 (changes requested):
I focused on pchain and highlighed some nits and some Blueberry related changes. Going to have a pass over it again

rajranjan0608:
> I focused on pchain and highlighed some nits and some Blueberry related changes. Going to have a pass over it again

Thanks for the review @abi87.

cc: @bilgehansahin for your visibility.

mapper/helper.go
```diff
@@ -11,3 +25,74 @@ func EqualFoldContains(arr []string, str string) bool {
...
+		hrp = constants.GetHRP(constants.MainnetID)
+	default:
+		return "", errUnrecognizedNetwork
+	}
+
+	return hrp, nil
+}
+
+// UnmarshalJSONMap converts map[string]interface{} into a interface{}.
+func UnmarshalJSONMap(m map[string]interface{}, i interface{}) error {
```
danlaine, line 45: I think there's very likely a better way to transform the information in `m` into an `i`. This is pretty error prone due to matching on exact spelling/casing of keys in the map / JSON tags on `i`.

mapper/pchain/tx_builder.go
```diff
@@ -0,0 +1,392 @@
...
+func buildImportTx(
+	matches []*parser.Match,
+	metadata Metadata,
+	codec codec.Manager,
+	avaxAssetID ids.ID,
+) (*txs.Tx, []*types.AccountIdentifier, error) {
+	blockchainID := metadata.BlockchainID
+	sourceChainID := metadata.SourceChainID
+
+	ins, imported, signers, err := buildInputs(matches[0].Operations, avaxAssetID)
```
danlaine, line 52: Are we guaranteed this index isn't OOB?

mapper/pchain/tx_builder.go
```diff
@@ -0,0 +1,392 @@
...
+	if err != nil {
+		return nil, nil, fmt.Errorf("parse inputs failed: %w", err)
+	}
+
+	outs, _, _, err := buildOutputs(matches[1].Operations, codec, avaxAssetID)
```
danlaine, line 57: Are we guaranteed this index isn't OOB?

mapper/pchain/tx_parser.go
```diff
@@ -0,0 +1,740 @@
...
+		txType = OpRemoveSubnetValidator
+		ops, err = t.parseRemoveSubnetValidatorTx(txID, unsignedTx)
+	case *txs.TransformSubnetTx:
+		txType = OpTransformSubnetValidator
+		ops, err = t.parseTransformSubnetTx(txID, unsignedTx)
+	case *txs.AdvanceTimeTx:
+		txType = OpAdvanceTime
+		// no op tx
+	default:
+		log.Printf("unknown type %T", unsignedTx)
```
danlaine, line 104: Should we return an error here?

mapper/pchain/tx_parser.go
```diff
@@ -0,0 +1,740 @@
...
+		return nil, err
+	}
+
+	var validator *pChainValidator.Validator
+	switch utx := rewardOuts.Tx.Unsigned.(type) {
+	case *txs.AddValidatorTx:
+		validator = &utx.Validator
+	case *txs.AddDelegatorTx:
+		validator = &utx.Validator
+	}
```
danlaine, line 259: Should we have a default case where we return an error?

mapper/pchain/tx_parser.go
```diff
@@ -0,0 +1,740 @@
...
+	addMetadataToStakeOuts(ops, validator)
+
+	return ops, nil
+}
+
+func addMetadataToStakeOuts(ops *txOps, validator *pChainValidator.Validator) {
+	if validator != nil {
```
danlaine, line 266: We can reduce indentation by inverting the condition:

```go
if validator == nil {
    return
}
...
```

mapper/pchain/tx_parser.go
```diff
@@ -0,0 +1,740 @@
...
+	if t.isConstruction {
+		status = nil
+	}
+
+	for _, in := range txIns {
+		metadata := &OperationMetadata{
+			Type: metaType,
+		}
+
+		if transferInput, ok := in.In.(*secp256k1fx.TransferInput); ok {
```
danlaine, line 327: What if this input is a `stakeable.LockIn` and not a `*secp256k1fx.TransferInput`?

mapper/pchain/tx_parser.go
```diff
@@ -0,0 +1,740 @@
...
+		}
+
+		// Rosetta cannot handle multisig at the moment. In order to pass data validation,
+		// we treat multisig outputs like a burn and inputs line a mint and therefore
+		// not include them in the operations
+		//
+		// Additionally, it is possible to have outputs without any addresses
+		// (e.g. https://testnet.avascan.info/blockchain/p/block/81016)
+		//
+		// therefore we skip parsing operations unless there is exactly 1 address
```
danlaine, line 411: I don't really know what the implications of this are but...is this OK? Is Rosetta going to report incorrect balances?

mapper/pchain/tx_parser.go
```diff
@@ -0,0 +1,740 @@
...
+		txIds = append(txIds, unsignedTx.TxID)
+	}
+
+	ids.SortIDs(txIds)
+
+	return txIds, nil
+}
+
+func getUniqueTxIds(ins []*avax.TransferableInput) []ids.ID {
+	txnIDs := make(map[string]ids.ID)
```
danlaine, line 624: Why don't we just keep the IDs in an `ids.Set`?

mapper/pchain/tx_parser.go
```diff
@@ -0,0 +1,740 @@
...
+
+	// Add reward UTXOs
+	for _, utxo := range d.RewardUTXOs {
+		utxos[utxo.OutputIndex] = utxo
+	}
+
+	return utxos
+}
+
+func mapUTXOs(txID ids.ID, outs []*avax.TransferableOutput, utxos map[uint32]*avax.UTXO) {
```
danlaine, line 669: Why do we take in a map rather than returning one?

mapper/pchain/tx_parser.go
```diff
@@ -0,0 +1,740 @@
...
+	}
+	return uniqueTxnIDs
+}
+
+func getUTXOMap(d *DependencyTx) map[uint32]*avax.UTXO {
+	utxos := make(map[uint32]*avax.UTXO)
+
+	if d.Tx != nil {
+		// Generate UTXOs from outputs
+		switch unsignedTx := d.Tx.Unsigned.(type) {
```
danlaine, line 641: Should we have a `default` case?

mapper/pchain/tx_parser_test.go
```diff
@@ -0,0 +1,293 @@
+package pchain
+
+import (
+	"testing"
+
+	"github.com/ava-labs/avalanchego/ids"
+	"github.com/ava-labs/avalanchego/utils/constants"
+	"github.com/ava-labs/avalanchego/vms/components/avax"
+	"github.com/coinbase/rosetta-sdk-go/types"
+	"github.com/stretchr/testify/assert"
```
danlaine, line 10: nit: we should probably use `"github.com/stretchr/testify/require"` instead -- it's semantically the same as `assert` but fails immediately on an error

client/client.go
```diff
@@ -36,10 +39,18 @@ type Client interface {
...
 	GetContractInfo(ethcommon.Address, bool) (string, uint8, error)
 	CallContract(context.Context, interfaces.CallMsg, *big.Int) ([]byte, error)
+	GetNetworkID(context.Context, ...rpc.Option) (uint32, error)
+	GetBlockchainID(context.Context, string, ...rpc.Option) (ids.ID, error)
+	IssueTx(ctx context.Context, txBytes []byte) (ids.ID, error)
+	GetAtomicUTXOs(ctx context.Context, addrs []string, sourceChain string, limit uint32, startAddress, startUTXOID string) ([][]byte, api.Index, error)
+	EstimateBaseFee(ctx context.Context) (*big.Int, error)
 }
 
+type EvmClient evm.Client
```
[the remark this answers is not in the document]
bilgehansahin: We were following the existing embedding pattern in `client.go`.   

`Client` is already used by the existing `info.Client` so aliasing allowed us to prevent the conflict.

client/client.go
```diff
@@ -36,10 +39,18 @@ type Client interface {
...
+
 type client struct {
 	info.Client
+	EvmClient
 	*EthClient
 	*ContractClient
```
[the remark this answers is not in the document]
bilgehansahin: ditto

client/pchainclient.go
```diff
@@ -0,0 +1,92 @@
...
+	GetNodeID(context.Context, ...rpc.Option) (ids.NodeID, error)
+	GetNetworkID(context.Context, ...rpc.Option) (uint32, error)
+	GetBlockchainID(context.Context, string, ...rpc.Option) (ids.ID, error)
+	GetTxFee(context.Context, ...rpc.Option) (*info.GetTxFeeResponse, error)
+}
+
+type (
+	indexerClient    = indexer.Client
+	platformvmClient = platformvm.Client
+	infoClient       = info.Client
```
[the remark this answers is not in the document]
bilgehansahin: same as C-chain client

mapper/pchain/tx_parser.go
bilgehansahin: It is correct if a particular account uses multisig. Rosetta operations are currently tied to 1 account. We also couldn't figure out how to model which account's balance we are counting a given UTXO against. 

With that in mind, this is the best trade-off we could think of short of making a Rosetta spec change.

service/backend/pchain/indexer/types.go
bilgehansahin: Not sure about the certificate. We used the indexer parser code shared by @rajranjan0608 . I will work with him offline to address this.

service/backend/pchain/backend.go
bilgehansahin: Is it safe to assume `txs.Version` will always match the `blocks.Codec`? I was hesitant to use it as they are exposed in different packages.

service/backend/common/parser.go
bilgehansahin: Initially planned to use for both p and c chain, but it didn't work out that way. Currently used by pchain block and indexer parser. Will copy it to the respective files considering it is few lines and we can use sane default values for version and codec.

service/backend/cchainatomictx/backend.go
bilgehansahin: We can address it in a future PR after coreth exports the codec value.

mapper/pchain/tx_parser.go
bilgehansahin: This function is called twice with different outs (Outs and StakeOuts) for the staking transactions. Passing the same map to both calls instead of dealing with merging the maps in the only caller of this.

mapper/pchain/tx_parser_test.go
bilgehansahin: Considering the existing tests also use `assert`, how about a separate PR to migrate all tests to use require instead?

mapper/pchain/tx_parser.go
bilgehansahin: Does ids.Set provide any benefits?

service/backend/pchain/indexer/types.go
abi87: to be clear, I am not saying it's wrong. I just realized there was a "missing" field and asked about the choice.

service/backend/pchain/backend.go
abi87: @bilgehansahin  thanks for the explanation.
TL;DR it is safe.
Let me provide some context. We have done quite some repackaging and cleanup of P-chain codebase which resulted (among other things) into splitting transactions and blocks into their own sub-packages.
The platformvm package's  codec is "blocks.Codec", its version is txs.Version. We may clean this up and introduce a blocks.Version variable to make more readable. Keep you posted on this

go.mod
```diff
@@ -42,6 +42,7 @@ require (
 	github.com/golang/mock v1.6.0 // indirect
 	github.com/golang/protobuf v1.5.2 // indirect
 	github.com/golang/snappy v0.0.4 // indirect
+	github.com/google/btree v1.1.2 // indirect
```
abi87, line 45: You should update avalanchego version to the latest 1.9.0 which includes relevant changes.
Also note that previous Blueberry name has been changed to Banff. We removed Blueberry from avalanchego codebase, you should do it too here.

service/backend/pchain/indexer/parser.go
```diff
@@ -0,0 +1,329 @@
...
+		parsedBlock.Txs = []*txs.Tx{castBlk.Tx}
+
+		// If the block has an advance time tx, use its timestamp as the block timestamp
+		for _, tx := range parsedBlock.Txs {
+			if att, ok := tx.Unsigned.(*txs.AdvanceTimeTx); ok {
+				blockTimestamp = att.Timestamp()
+				break
+			}
+		}
+	case *blocks.BlueberryProposalBlock:
```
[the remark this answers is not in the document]
bilgehansahin: ack

service/backend/pchain/indexer/parser.go
```diff
@@ -0,0 +1,329 @@
...
+		errs.Add(fmt.Errorf("no handler exists for block type %T", castBlk))
+	}
+
+	if blockTimestamp.IsZero() {
+		if proposer.Timestamp > genesisTimestamp {
+			blockTimestamp = time.Unix(proposer.Timestamp, 0)
+		} else {
+			blockTimestamp = time.Unix(genesisTimestamp, 0)
+		}
+	}
```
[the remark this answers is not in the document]
bilgehansahin: The idea here is to return the block timestamp if one was found; if not, then return proposer level timestamp. I wasn't sure if either has to exist, so I'm falling back to the genesis timestamp as the default.

It is technically incorrect since we simply don't have a clue about the timestamp, but if we return 0 instead, then the Rosetta validation complains about it. I'll add a comment to explain the context.

go.mod
abi87: So mainnet is under deployment today (see release [here](https://github.com/ava-labs/avalanchego/releases/tag/v1.9.0)), so it should be safe to merge, unless your workflow has reasons I am not aware of.

service/backend/pchain/block.go
```diff
@@ -0,0 +1,445 @@
...
+	genesisBlock, err := b.getGenesisBlock(ctx)
+	if err != nil {
+		return false, err
+	}
+
+	if hash != "" {
+		return hash == genesisBlock.BlockID.String(), nil
+	}
+
+	return index == int64(genesisBlock.Height), nil
```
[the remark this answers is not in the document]
bilgehansahin: The checks for hash and index are separated because in block or transaction lookup requests, one can either pass index or hash. If hash is present, we compare hash with genesis hash. If it is omitted, we then check the height. 

This line would default to true if a non-genesis block is looked up by hash only. I will add comments to clarify this.

service/backend/pchain/indexer/parser.go
```diff
@@ -0,0 +1,329 @@
...
+	return p.ParseBlockAtIndex(ctx, height)
+}
+
+func (p *parser) ParseBlockAtIndex(ctx context.Context, index uint64) (*ParsedBlock, error) {
+	err := p.initCtx(ctx)
+	if err != nil {
+		return nil, err
+	}
+
+	container, err := p.pChainClient.GetContainerByIndex(ctx, index-1)
```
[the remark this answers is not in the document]
bilgehansahin: I'm not sure if I understand what you are asking. Container indices seem to be off by one compared to block heights in P-chain (e.g. block 10 seems to be indexed in container id 9) and hence we have special handling for the genesis block. I need to perform the -1 somewhere, are you suggesting we should do it in the clients of this method instead of here?

mapper/cchainatomictx/tx_builder.go
```diff
@@ -0,0 +1,163 @@
...
+		BlockchainID:   metadata.CChainID,
+		SourceChain:    *metadata.SourceChainID,
+		ImportedInputs: importedInputs,
+		Outs:           outs,
+	}}
+	return tx, signers, nil
+}
+
+func buildIns(matches []*parser.Match, metadata Metadata, avaxAssetID ids.ID) ([]evm.EVMInput, []*types.AccountIdentifier) {
+	inputMatch := matches[0]
```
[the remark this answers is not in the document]
bilgehansahin: Yes, the method converting request to matches ensures that.

mapper/cchainatomictx/tx_builder.go
```diff
@@ -0,0 +1,163 @@
...
+			AssetID: avaxAssetID,
+			Nonce:   metadata.Nonce,
+		})
+		signers = append(signers, op.Account)
+	}
+	return ins, signers
+}
+
+func buildImportedInputs(matches []*parser.Match, avaxAssetID ids.ID) ([]*avax.TransferableInput, []*types.AccountIdentifier, error) {
+	inputMatch := matches[0]
```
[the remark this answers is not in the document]
bilgehansahin: Yes

mapper/cchainatomictx/tx_builder.go
```diff
@@ -0,0 +1,163 @@
...
+			return nil, nil, err
+		}
+
+		importedInputs = append(importedInputs, &avax.TransferableInput{
+			UTXOID: *utxoID,
+			Asset:  avax.Asset{ID: avaxAssetID},
+			In: &secp256k1fx.TransferInput{
+				Amt: inputMatch.Amounts[i].Uint64(),
+				Input: secp256k1fx.Input{
+					SigIndices: []uint32{0},
```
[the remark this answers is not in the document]
bilgehansahin: We don't support multisig construction so at the moment, yes.

mapper/cchainatomictx/tx_builder.go
```diff
@@ -0,0 +1,163 @@
...
+		})
+		signers = append(signers, op.Account)
+	}
+	avax.SortTransferableInputs(importedInputs)
+
+	return importedInputs, signers, nil
+}
+
+func buildOuts(matches []*parser.Match, avaxAssetID ids.ID) []evm.EVMOutput {
+	outputMatch := matches[1]
```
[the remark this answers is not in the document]
bilgehansahin: Yes

mapper/cchainatomictx/tx_builder.go
```diff
@@ -0,0 +1,163 @@
...
+
+func buildIns(matches []*parser.Match, metadata Metadata, avaxAssetID ids.ID) ([]evm.EVMInput, []*types.AccountIdentifier) {
+	inputMatch := matches[0]
+
+	ins := []evm.EVMInput{}
+	signers := []*types.AccountIdentifier{}
+	for i, op := range inputMatch.Operations {
+		ins = append(ins, evm.EVMInput{
+			Address: ethcommon.HexToAddress(op.Account.Address),
+			Amount:  inputMatch.Amounts[i].Uint64(),
```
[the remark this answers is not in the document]
bilgehansahin: Yes, per https://github.com/coinbase/rosetta-sdk-go/blob/1cc6016f48cbb1ac3113f3b584a0cf3a83063b6d/parser/match_operations.go#L672, Amounts and Operations have the same length. We also ensure they are not nil as part of the operation parsing.

mapper/cchainatomictx/tx_builder.go
```diff
@@ -0,0 +1,163 @@
...
+		utxoID, err := mapper.DecodeUTXOID(op.CoinChange.CoinIdentifier.Identifier)
+		if err != nil {
+			return nil, nil, err
+		}
+
+		importedInputs = append(importedInputs, &avax.TransferableInput{
+			UTXOID: *utxoID,
+			Asset:  avax.Asset{ID: avaxAssetID},
+			In: &secp256k1fx.TransferInput{
+				Amt: inputMatch.Amounts[i].Uint64(),
```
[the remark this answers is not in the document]
bilgehansahin: Yes

mapper/cchainatomictx/tx_builder.go
```diff
@@ -0,0 +1,163 @@
...
+			Amount:  outputMatch.Amounts[i].Uint64(),
+			AssetID: avaxAssetID,
+		})
+	}
+
+	return outs
+}
+
+func buildExportedOutputs(matches []*parser.Match, codec codec.Manager, avaxAssetID ids.ID) ([]*avax.TransferableOutput, error) {
+	outputMatch := matches[1]
```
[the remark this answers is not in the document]
bilgehansahin: Yes

mapper/helper.go
bilgehansahin: We simply moved the existing private function from service/helper.go to here as-is for reuse in the new code. We can look at refactoring the mechanism in a future PR.

mapper/pchain/tx_builder.go
bilgehansahin: Yes

mapper/pchain/tx_builder.go
bilgehansahin: Yes

mapper/pchain/tx_builder.go
```diff
@@ -0,0 +1,392 @@
...
+	codec codec.Manager,
+	avaxAssetID ids.ID,
+) (*txs.Tx, []*types.AccountIdentifier, error) {
+	if metadata.ExportMetadata == nil {
+		return nil, nil, errInvalidMetadata
+	}
+	blockchainID := metadata.BlockchainID
+	destinationChainID := metadata.DestinationChainID
+
+	ins, _, signers, err := buildInputs(matches[0].Operations, avaxAssetID)
```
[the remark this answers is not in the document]
bilgehansahin: Yes

mapper/pchain/tx_builder.go
```diff
@@ -0,0 +1,392 @@
...
+	if err != nil {
+		return nil, nil, fmt.Errorf("parse inputs failed: %w", err)
+	}
+
+	outs, _, exported, err := buildOutputs(matches[1].Operations, codec, avaxAssetID)
```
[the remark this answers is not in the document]
bilgehansahin: Yes

mapper/pchain/tx_parser.go
bilgehansahin: No, we don't necessarily want to block parsing if we encounter new transaction types that's not supported yet in Rosetta.

go.mod
bilgehansahin: done

mapper/cchainatomictx/tx_builder.go
```diff
@@ -0,0 +1,163 @@
...
+	for i, op := range inputMatch.Operations {
+		ins = append(ins, evm.EVMInput{
+			Address: ethcommon.HexToAddress(op.Account.Address),
+			Amount:  inputMatch.Amounts[i].Uint64(),
+			AssetID: avaxAssetID,
+			Nonce:   metadata.Nonce,
+		})
+		signers = append(signers, op.Account)
+	}
+	return ins, signers
```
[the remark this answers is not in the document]
bilgehansahin: I'm not sure if we need it, but will sort them just in case.

service/backend/pchain/backend.go
```diff
@@ -0,0 +1,111 @@
...
+	assetID ids.ID,
+	networkIdentifier *types.NetworkIdentifier,
+) *Backend {
+	return &Backend{
+		fac:               &crypto.FactorySECP256K1R{},
+		networkIdentifier: networkIdentifier,
+		pClient:           pClient,
+		getUTXOsPageSize:  1024,
+		codec:             blocks.Codec,
+		codecVersion:      txs.Version,
```
abi87, line 52: TL;DR 
```
codec:        blocks.Codec,
codecVersion: blocks.Version,
```

nit: we have cleaned up platformvm to ease up codecs usage. Here you can now only use blocks.Codec/Version and forget about txs.Codec

service/backend/pchain/indexer/parser.go
```diff
@@ -0,0 +1,340 @@
...
+
+func NewParser(pChainClient client.PChainClient) (Parser, error) {
+	errs := wrappers.Errs{}
+
+	aliaser := ids.NewAliaser()
+	errs.Add(aliaser.Alias(constants.PlatformChainID, mapper.PChainNetworkIdentifier))
+
+	return &parser{
+		codec:            pBlocks.Codec,
+		codecVersion:     txs.Version,
```
abi87, line 64: see previous comment
```
codec:        blocks.Codec,
codecVersion: blocks.Version,
```

service/backend/pchain/indexer/parser.go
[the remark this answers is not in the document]
abi87: Let me cross check on avalanche go side. May be a behaviour I missed (maybe genesis is not indexed)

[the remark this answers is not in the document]
abi87: You are right. So index API does not index genesis block. So height n maps to index n-1.
Thanks for bringing it up

service/backend/pchain/backend.go
bilgehansahin: Awesome, will push an update with this today.

client/pchainclient.go
```diff
@@ -0,0 +1,93 @@
...
+
+	GetAssetDescription(ctx context.Context, assetID string, options ...rpc.Option) (*avm.GetAssetDescriptionReply, error)
+
+	// info.Client methods
+	IsBootstrapped(context.Context, string, ...rpc.Option) (bool, error)
+	Peers(context.Context, ...rpc.Option) ([]info.Peer, error)
+	GetNodeID(context.Context, ...rpc.Option) (ids.NodeID, *signer.ProofOfPossession, error)
+	GetNetworkID(context.Context, ...rpc.Option) (uint32, error)
+	GetBlockchainID(context.Context, string, ...rpc.Option) (ids.ID, error)
+	GetTxFee(context.Context, ...rpc.Option) (*info.GetTxFeeResponse, error)
```
joshua-kim, line 63: 1. Why not embed interfaces?
2. Maybe this is turning into a god-interface? I don't think the p-chain client interface should have access to info api/etc.

joshua-kim (changes requested):
It's not really clear to me what this PR's changes are for (what behavior is changed/not/why/etc), and I think it's mostly due to the sheer size of the diff (17k added/250 removed). Because of this, I didn't take a deep look into this PR because I felt I couldn't provide a useful review in its current state.

At the bare minimum, can we post in the body of this PR what behavior is changed/not/etc and why these changes are needed?

Ideally, to increase the pace + quality at which this diff can be merged would be to break up the PR into logical components; as an example I think it's strange that both the client + backend interface definition + implementations are all wrapped together in a single pr. That might be one example of an area we could split this to review this more easily.
  - Good article i read on why small PRs are preferable [here](https://www.swarmia.com/blog/why-small-pull-requests-are-better/)

Also, how was this tested? Can we add testing details to the PR description?

service/backend/pchain/account_test.go
```diff
@@ -0,0 +1,376 @@
...
+				},
+				Currencies: []*types.Currency{
+					mapper.AtomicAvaxCurrency,
+				},
+			},
+		)
+
+		assert.Nil(t, resp)
+		assert.Equal(t, "Internal server error", err.Message)
+		assert.Equal(t, "new block added while fetching utxos", err.Details["error"])
```
joshua-kim, line 182: Can we assert the underlying wrapped error type w/ `errors.Is` instead of just parsing the error text?

service/backend/pchain/indexer/parser.go
```diff
@@ -0,0 +1,340 @@
...
+	"github.com/ava-labs/avalanchego/vms/platformvm/txs"
+	proposerBlk "github.com/ava-labs/avalanchego/vms/proposervm/block"
+
+	"github.com/ava-labs/avalanche-rosetta/client"
+	"github.com/ava-labs/avalanche-rosetta/mapper"
+)
+
+var errParserUninitialized = errors.New("uninitialized parser")
+
+const genesisTimestamp = 1599696000
```
joshua-kim, line 28: Can we use a date and convert it to its timestamp representation instead to make this more verbose

postman/test_signing_server.go
```diff
@@ -0,0 +1,145 @@
...
+type signRequest struct {
+	Payloads []*types.SigningPayload `json:"payloads"`
+}
+
+type signResponse struct {
+	Signatures []*types.Signature `json:"signatures"`
+}
+
+type signingServer struct {
+	port       uint
```
joshua-kim, line 25: this breaks the tcp spec because it allows invalid port numbers; port numbers are only 16 bits

ref: https://www.ietf.org/rfc/rfc793.txt

```
    0                   1                   2                   3   
    0 1 2 3 4 5 6 7 8 9 0 1 2 3 4 5 6 7 8 9 0 1 2 3 4 5 6 7 8 9 0 1 
   +-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+
   |          Source Port          |       Destination Port        |
   +-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+
   |                        Sequence Number                        |
   +-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+
   |                    Acknowledgment Number                      |
   +-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+
   |  Data |           |U|A|P|R|S|F|                               |
   | Offset| Reserved  |R|C|S|S|Y|I|            Window             |
   |       |           |G|K|H|T|N|N|                               |
   +-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+
   |           Checksum            |         Urgent Pointer        |
   +-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+
   |                    Options                    |    Padding    |
   +-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+
   |                             data                              |
   +-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+

                            TCP Header Format

          Note that one tick mark represents one bit position.

                               Figure 3.

  Source Port:  16 bits

    The source port number.

  Destination Port:  16 bits

    The destination port number.
```

postman/test_signing_server.go
```diff
@@ -0,0 +1,145 @@
...
+
+	return &signingServer{
+		privateKey: privateKey,
+		publicKey:  publicKey,
+		port:       port,
+	}
+}
+
+var opts struct {
+	port       uint
```
joshua-kim, line 130: same tcp spec comment here ^

service/backend/common/construction.go
```diff
@@ -0,0 +1,347 @@
...
+			},
+		},
+		ErrUnmatched: true,
+	}
+
+	return parser.MatchOperations(descriptions, operations)
+}
+
+type TxBuilder interface {
+	BuildTx(matches []*types.Operation, rawMetadata map[string]interface{}) (AvaxTx, []*types.AccountIdentifier, *types.Error)
```
joshua-kim, line 108: nit: idiomatic to use `any` in place of `interface{}` in go1.18+

nicolas-martin:
@joshua-kim 

If you look at the commits you'll find all the smaller PR previously included and explained. ie(https://github.com/ava-labs/avalanche-rosetta/pull/122)

joshua-kim:
> @joshua-kim
> If you look at the commits you'll find all the smaller PR previously included and explained. ie(https://github.com/ava-labs/avalanche-rosetta/pull/122)

Didn't realize this was our feature branch and these were previously approved, thanks for the context 👍 .

postman/test_signing_server.go
```diff
@@ -0,0 +1,145 @@
...
+	flag.StringVar(&opts.privateKey, "private-key", "", "private key to sign with")
+	flag.Parse()
+}
+
+func main() {
+	println("!!! DO NOT USE THIS UTILITY WITH PRIVATE KEYS THAT HAS ACCESS TO REAL FUNDS!!!\n")
+	println("This is a test utility used to help testing Rosetta AVAX implementation during development\n\n")
+	signingServer := newSigningServer(opts.port, opts.privateKey)
+	signingServer.run()
+}
```
joshua-kim, line 145: Should this be removed?

service/backend/common/account.go
```diff
@@ -0,0 +1,32 @@
...
+
+	"github.com/coinbase/rosetta-sdk-go/types"
+)
+
+// SortUnique deduplicates given slice of coins and sorts them by UTXO id in ascending order for consistency
+//
+// Per https://docs.avax.network/apis/avalanchego/apis/p-chain#platformgetutxos
+// and https://docs.avax.network/apis/avalanchego/apis/c-chain#avaxgetutxos
+// paginated getUTXOs calls may have duplicate UTXOs in different pages, this helper eliminates them.
+func SortUnique(coins []*types.Coin) []*types.Coin {
```
joshua-kim, line 14: I think this can be replaced with the new 1.18 features `slices.Sort` + `slices.Compact` (might need to implement some interfaces in `types.Coin`)

service/backend/common/construction.go
```diff
@@ -0,0 +1,347 @@
...
+
+var (
+	errNoOperationsToMatch      = errors.New("no operations were passed to match")
+	errInvalidInput             = errors.New("invalid input")
+	errInvalidInputSignatureLen = errors.New("input signature length doesn't match credentials needed")
+	errInsufficientSignatures   = errors.New("insufficient signatures")
+	errInvalidSignatureLen      = errors.New("invalid signature length")
+)
+
+func DeriveBech32Address(fac *crypto.FactorySECP256K1R, chainIDAlias string, req *types.ConstructionDeriveRequest) (*types.ConstructionDeriveResponse, *types.Error) {
```
joshua-kim, line 31: We should add comments to anything that's exported (what the function does, invariants, etc)

.golangci.yml
```diff
@@ -8,6 +8,8 @@ run:
 issues:
   # Maximum count of issues with the same text. Set to 0 to disable. Default is 3.
   max-same-issues: 0
+  exclude:
+    - "G114: Use of net/http serve function that has no support for setting timeouts"
```
joshua-kim, line 12: question: why do we need this override?

client/pchainclient.go
```diff
@@ -0,0 +1,93 @@
...
+
+type pchainClient struct {
+	platformvmClient
+	indexerClient
+	infoClient
+	xChainClient avm.Client
+}
+
+// NewPChainClient returns a new client for Avalanche APIs related to P-chain
+func NewPChainClient(ctx context.Context, endpoint, indexerEndpoint string) PChainClient {
```
joshua-kim, line 80: in-line with the previous comment this wraps more than just the p-chain

cmd/server/config.go
```diff
@@ -61,6 +63,10 @@ func (c *config) ApplyDefaults() {
 		c.RPCEndpoint = "http://localhost:9650"
 	}
 
+	if c.IndexerEndpoint == "" {
+		c.IndexerEndpoint = c.RPCEndpoint
```
joshua-kim, line 67: Does this work? Our index apis have a specific endpoint which is distinct from our rpc endpoints (ref: https://docs.avax.network/apis/avalanchego/apis/index-api#endpoints)

client/pchainclient.go
```diff
@@ -0,0 +1,93 @@
...
+	endpoint = strings.TrimSuffix(endpoint, "/")
+
+	return pchainClient{
+		platformvmClient: platformvm.NewClient(endpoint),
+		xChainClient:     avm.NewClient(endpoint, "X"),
+		infoClient:       info.NewClient(endpoint),
+		indexerClient:    indexer.NewClient(indexerEndpoint + "/ext/index/P/block"),
```
joshua-kim, line 87: Related to the later comment on how this flag is passed, but this looks like `indexerEndpoint` is basically expected to just be the raw endpoint. I think it would make sense to either use the `endpoint` instead or to have `indexerEndpoint` be expected to already have this suffix.

client/pchainclient.go
bilgehansahin: We are mimicking what the existing client used for C-chain does here to stay consistent as much as possible.

cmd/server/config.go
bilgehansahin: see below

client/pchainclient.go
bilgehansahin: Having `indexerEndpoint` separate from `endpoint` allows us to point to public indexer endpoint while pointing endpoint to the public api endpoint which are separate.

Since the suffix can potentially be changed in future versions as part of a node update, I think having it here instead of let it come from the configuration would allow silently changing it in code instead of needing users to update their configs.

.golangci.yml
bilgehansahin: Without it, the `http.ListenAndServe` call in main.go would break linting. This rule was added to golangcilint while we are working on this change and fixing the underlying issue is outside the scope of this change. In the future, a separate PR can update the server startup and this rule can be removed.

service/backend/common/account.go
bilgehansahin: When we started the work on this repo, it was only 1.17 compatible and we have written our changes with that. I think a golang 1.18+ refactoring can be done along with the rest of the code in a separate PR.

postman/test_signing_server.go
bilgehansahin: This is a test utility that I used together with the postman file attached. It allows one to execute the full construction flow through postman. You can see the README in the same folder on how it is used. 

I added it here based on @rajranjan0608's request as it helps with e2e testing during development. If the team things we shouldn't add, I can remove it.

client/pchainclient.go
bilgehansahin: It is named PChainClient  to differentiate from the existing `Client` that is used for C-chain operations but includes info as well. I agree we could probably be more creative with the name but for consistency with `Client` kept it as this. Here the name mainly refers to the usage of this type which is fulfilling P-chain specific needs.

service/backend/pchain/indexer/parser.go
bilgehansahin: Sure

postman/test_signing_server.go
bilgehansahin: I'm familiar with the spec. Since flag does not have a uint16 option out of the box and passing a large value errors the startup with `listen tcp: address <address>: invalid port` which is descriptive enough for a test utility, I didn't bother with checking limits / converting to uint16.

service/backend/common/construction.go
bilgehansahin: Sure, I can add some comments.

postman/test_signing_server.go
rajranjan0608: Signing helps in testing some of the endpoints through Postman. Should have no problem since we called it out in CAPS.

.golangci.yml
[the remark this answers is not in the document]
bilgehansahin: This is a new check added to `golangci-lint` while we are working on this change over last 3 months. Fixing the code causing this is beyond the scope of this PR. I disabled the rule in the mean time.

Also see https://github.com/ava-labs/avalanche-rosetta/pull/123#discussion_r993896755

cmd/server/main.go
```diff
@@ -111,17 +116,31 @@ func main() {
...
 
-	network := &types.NetworkIdentifier{
+	networkP := &types.NetworkIdentifier{
+		Blockchain: service.BlockchainName,
+		Network:    cfg.NetworkName,
+		SubNetworkIdentifier: &types.SubNetworkIdentifier{
+			Network: mapper.PChainNetworkIdentifier,
+		},
+	}
+	networkC := &types.NetworkIdentifier{
```
abi87, line 126: should probably be:
```
networkC := &types.NetworkIdentifier{
	Blockchain: service.BlockchainName,
	Network:    cfg.NetworkName,
	SubNetworkIdentifier: &types.SubNetworkIdentifier{
		Network: mapper.CChainNetworkIdentifier,
	},
}
```

bilgehansahin: SubNetworkIdentifier field for C-chain was omitted deliberately for backwards compatibility with existing consumers of Rosetta who makes requests without specifying the SubNetworkIdentifier field for C-chain. IIRC, suggestion came from @patrick-ogrady.

abi87: Thanks for the context, @bilgehansahin

service/backend/pchain/indexer/parser.go
abi87: @bilgehansahin we are pulling blocks from avalanchego index API, which provide a timestamp for the block. Why don't we use that?

service/backend/pchain/indexer/parser.go
```diff
@@ -0,0 +1,349 @@
...
+		}
+	}
+	parsedBlock.Timestamp = blockTimestamp.UnixMilli()
+
+	return &parsedBlock, errs.Err
+}
+
+func getProposerFromBytes(bytes []byte) (Proposer, []byte, error) {
+	proposer, _, err := proposerBlk.Parse(bytes)
+	if err != nil || proposer == nil {
```
abi87, line 314: @bilgehansahin why is this error silenced? We really should further process bytes if parser errored

bilgehansahin: I have copied this function verbatim from the indexer parsing reference implementation we were provided. I'm not very familiar with the how proposer works. We can handle it differently if you have a suggestion.

abi87: I looked into this and it is correct. Briefly, we have to handle blocks before and after snowman++ activation. To do that we try parsing a block as proposerVM block and if it fails, we try and a simple pre-snowman++ Pchain block.

In fact I have refactored this for clarity in the [Indexer cleanup PR](https://github.com/ava-labs/avalanche-rosetta/pull/160). Would really appreciate your input there

service/backend/pchain/indexer/parser.go
bilgehansahin: AFAICT, the container timestamp refers to the time a given container is accepted by the indexer node, not the time that block was created or broadcasted. For example in my test node I have been using since June, all containers that was published before the node was started gets a timestamp of the initial sync of the node in June. 

We wanted the timestamps to be consistent across nodes / deployments and hence looked at the block time. Pre-Banff it was a hit-and-miss, so if a block had a missing timestamp we injected the so called genesis timestamp as timestamps are required by Rosetta spec and the validation requires a timestamp after year 2000. With Banff, it seems every block has a timestamp so we can get an accurate timestamp as far as I can tell.

abi87: @bilgehansahin, you're right about container timestamp limitation; you are also right that for post Banff blocks, the block timestamp wil be the `chain Time` at which that block is generated, hence that time is guaranteed to be monotonic and agreed upon by any node in the network.
The reason I proposed the indexer timestamp alternative is that indexer guarantee a monotonic time pre-Banff and across Banff fork activation. On the contrary in the current solution, we cut and past times from different source with no guarantee of monotonicity.
I am not sure what are the business requirements here and I am favour of picking the solution that serve them the best.

mapper/pchain/tx_parser.go
```diff
@@ -0,0 +1,883 @@
...
+		case *txs.CreateChainTx:
+			mapUTXOs(d.Tx.ID(), unsignedTx.Outs, utxos)
+		default:
+			// no utxos extracted from unsupported transaction types
+		}
+	}
+
+	// Add reward UTXOs
+	for _, utxo := range d.RewardUTXOs {
+		utxos[utxo.OutputIndex] = utxo
```
abi87, line 806: @bilgehansahin I doubt this is correct; I think this op may overwrite some of the utxos previously filled.
Call `A` the tx whose dependency are represented by `d` and call `B` `d.Tx` i.e. the tx carrying the utxos spent by `A`
We fill `utxos` with two data:

1. all utxos from `B`
2. RewardUTXO which is built by avalanchego as a utxo from `A` with index len(A.utxos)+len(A.stakes).

So if B has many more utxos than A, this op would overwrite the a B utxo.
I believe we need to handle these dependencies totally differently

mapper/types.go
```diff
@@ -60,6 +71,11 @@ var (
 		Decimals: 18,
 	}
 
+	AtomicAvaxCurrency = &types.Currency{
```
abi87, line 74: @bilgehansahin I don't quite get why we have this. Why can't we just use `AvaxCurrency` defined above?

mapper/pchain/tx_parser.go
bilgehansahin: Sure 👍

mapper/types.go
bilgehansahin: We use this to represent amounts in nanoAVAX (see the `decimals: 9` below and 18 in AvaxCurrency above). It makes it easier to use Rosetta for non C-chain transactions rather than converting everyting into weis back and forth.

mapper/pchain/tx_parser.go
abi87: Indeed I have opened the PR https://github.com/ava-labs/avalanche-rosetta/pull/168 to fix this. @bilgehansahin please have a look

service/backend/pchain/construction.go
```diff
@@ -0,0 +1,337 @@
...
+
+	var suggestedFee *types.Amount
+	var metadata *pmapper.Metadata
+	switch opMetadata.Type {
+	case pmapper.OpImportAvax:
+		metadata, suggestedFee, err = b.buildImportMetadata(ctx, req.Options)
+	case pmapper.OpExportAvax:
+		metadata, suggestedFee, err = b.buildExportMetadata(ctx, req.Options)
+	case pmapper.OpAddValidator, pmapper.OpAddDelegator:
+		metadata, suggestedFee, err = b.buildStakingMetadata(req.Options)
```
abi87, line 71: @bilgehansahin are these the only tx Rosetta can build? Why?

abi87 (commented):
@bilgehansahin I merged the PRs I had opened and opened a couple of other minor ones fixing bugs and cleaning up.
I did not find other bugs than the one solved in the PRs I mentioned.
I guess there are two outstanding points before we can be content with code correctness and approve this PR:

1.  Can you rerun whatever e2e you had with the changes? We should add some more UTs too, (especially considering the lock inputs left out) but at the very list we should resync to check
2. On construction side, it's kind of confusing what actual transactions Rosetta is capable of building. Can you clarify

As we agree, any other code cleanup will be delaying to next PRs

mapper/types.go
bilgehansahin: We use AtomicAvaxCurrency in P-chain and C-chain atomic transactions. AvaxCurrency should only be used in C-chain transactions.

service/backend/pchain/construction.go
bilgehansahin: We have added the transactions necessary for staking AVAX. We left the remaining transactions for future updates.

nicolas-martin:
@abi87 Seems like cchain get `/block` by ID is broken
```
curl --request POST \
  --url http://localhost:9090/block \
  --header 'Content-Type: application/json' \
  --data '{
        "network_identifier": {
                "blockchain": "Avalanche",
                "network": "Fuji"
        },
        "block_identifier": {
                "index":2
        }
}'
{"code":7,"message":"Client error","retriable":true,"details":{"error":"processing block 1 failed: engine finalization check failed: invalid block due to failed semanatic verify: failed to fetch import UTXOs from 2JVSBoinj9C2J33VntvzYtVJNZdN2NKiwwKjcumHUWEb5DbBrm due to: not found at height 1"}}
```
Not sure why it's complaining about UTXO in this case. Also typo in `semanatic`. How are you testing these changes?

nicolas-martin:
@abi87 
I'm getting an error fetch cchain blocks by height
```
curl --request POST \
  --url http://localhost:9090/block \
  --header 'Content-Type: application/json' \
  --data '{
        "network_identifier": {
                "blockchain": "Avalanche",
                "network": "Fuji"
        },
        "block_identifier": {
                "index":2
        }
}'
{"code":7,"message":"Client error","retriable":true,"details":{"error":"processing block 1 failed: engine
 finalization check failed: invalid block due to failed semanatic verify: failed to fetch import UTXOs 
from 2JVSBoinj9C2J33VntvzYtVJNZdN2NKiwwKjcumHUWEb5DbBrm due to: not found at height 1"}}
```

How are you testing these changes?

abi87 (approved):

learyce (approved):

PR #158, opened 2022-10-10
[Chore] Fix exported outputs' operation identifier

xiaoxiaff:
Fix nil operation identifier issue.


xiaoxiaff:
Change has been merged to #157

PR #162, opened 2022-10-20
Backend block cleanup

abi87:


service/backend/pchain/block.go
```diff
@@ -299,62 +285,13 @@ func (b *Backend) getChainIDs(ctx context.Context) (map[string]string, error) {
 	return b.chainIDs, nil
 }
 
-func (b *Backend) getBlockDetails(ctx context.Context, index int64, hash string) (*indexer.ParsedBlock, error) {
```
abi87, line 302: `getBlockDetails` makes two network calls:

1. first retrieve full block from pchain api just to pick its height
2. then with the height retrieved block and timestamp from indexer api

Just refactored this function to make a single network call

service/backend/pchain/block.go
```diff
@@ -299,62 +285,13 @@ func (b *Backend) getChainIDs(ctx context.Context) (map[string]string, error) {
...
 	}
 
-	// if hash is provided, make sure it matches genesis block hash
-	if hash != "" {
-		return hash == genesisBlock.BlockID.String(), nil
-	}
-
-	// if hash is omitted, check if the height matches the genesis block height
-	return index == int64(genesisBlock.Height), nil
+	return hash == genesisBlock.BlockID.String() || index == int64(genesisBlock.Height), nil
```
abi87, line 294: I still don't see how this change is not equivalent no the first one

service/backend/pchain/block.go
```diff
@@ -421,27 +358,12 @@ func (b *Backend) buildGenesisAllocationTx() (*txs.Tx, error) {
...
 	allocationTx := &txs.ImportTx{}
 	allocationTx.Outs = outs
 	tx := &txs.Tx{
 		Unsigned: allocationTx,
 	}
-	return tx, nil
-}
-
-// initializes tx to have tx identifier generated
-func (b *Backend) initializeTx(tx *txs.Tx) error {
```
abi87, line 433: just use the right methods from platformvm

service/backend/pchain/block.go
bilgehansahin: Requests for /block can specify either height or hash. If a request only specifies a non-genesis hash, the height will be defaulted to 0, and the index passed down to here will be 0 and will match the genesis height, even though the request is for a hash that is not for the genesis block. 

If you really want to write this in a single line, you probably should do something like:

```
return hash == genesisBlock.BlockID.String() || hash == "" && index == int64(genesisBlock.Height), nil
```

service/backend/pchain/block.go
```diff
@@ -421,27 +358,12 @@ func (b *Backend) buildGenesisAllocationTx() (*txs.Tx, error) {
...
-
-	signedBytes, err := b.codec.Marshal(b.codecVersion, tx)
-	if err != nil {
-		return err
-	}
-
-	tx.Initialize(unsignedBytes, signedBytes)
-
-	return nil
+	return tx, tx.Sign(blocks.Codec, nil)
```
bilgehansahin, line 368: You can use `b.codec` instead.

client/client.go
```diff
@@ -19,7 +19,13 @@ import (
 var _ Client = &client{}
 
 type Client interface {
+	// info.Client methods
```
abi87, line 22: just a minor methods reordering by client

cmd/server/main.go
```diff
@@ -57,7 +58,7 @@ func main() {
 		log.Fatal("config validation error:", err)
 	}
 
-	apiClient, err := client.NewClient(context.Background(), cfg.RPCBaseURL)
+	cChainClient, err := client.NewClient(context.Background(), cfg.RPCBaseURL)
```
abi87, line 61: just a renaming for readability.

service/backend/pchain/backend.go
```diff
@@ -104,3 +107,24 @@ func (b *Backend) getGenesisBlock(ctx context.Context) (*indexer.ParsedGenesisBl
 	}
 	return genesisBlock, nil
 }
+
+func (b *Backend) initChainIDs() error {
```
abi87, line 111: moved getChainIDs to backend constructor

service/backend/pchain/block.go
abi87: Thanks for the patient explaination, I could not see the default case. Rolling back you initial implementation.

service/backend/pchain/block.go
abi87: Done

service/backend/pchain/backend.go
bilgehansahin: [Rosetta has the concept of an offline mode](https://www.rosetta-api.org/docs/construction_api_introduction.html#completely-offline), where all construction endpoints except `/construction/metadata` must be able to run without connection to the node. Because of this, we cannot put client calls in the backend constructor by default. We lazy loaded the mappings with caching for this reason.

abi87: Thanks for clarifying this. I undestand node mode cannot change at run time. Hence I fixed this by passing service mode to Backend constructor.

abi87: Thanks for the explainantion. Since service mode cannot change at run time, we can still build chainsID mapping at backend construction, just passing in service mode.

bilgehansahin (approved):

PR #163, opened 2022-10-21
Pchain network id checks

abi87:
Network name (i.e. "mainnet" "fuji" "local"...) is a static information configured at rosetta startup but never really checked in requets. This PR introduces two checks and a minor thing:

1. Upon Rosetta startup, we check whether configured network name matches the one supported by avalanchego node. If not Rosetta won't start.
2. Upon every request, request network name is compared with configured one. Request is only served if they matches.
3. Every quantity depending on network name only (e.g. network HRP) is prefetched upon startup


mapper/pchain/helper.go
```diff
@@ -1,14 +0,0 @@
-package pchain
-
-import (
-	"github.com/coinbase/rosetta-sdk-go/types"
-
-	"github.com/ava-labs/avalanche-rosetta/mapper"
-)
-
-// IsPChain checks network identifier to make sure sub-network identifier set to "P"
-func IsPChain(networkIdentifier *types.NetworkIdentifier) bool {
```
abi87, line 10: just moved to pchain backend, only caller

cmd/server/main.go
```diff
@@ -196,6 +187,32 @@ func main() {
...
+		return fmt.Errorf("cant fetch network name: %w", err)
+	}
+
+	if cfg.NetworkName == "" {
+		log.Printf("network name is not provided, set to %s", networkName)
+		cfg.NetworkName = networkName
+		return nil
+	}
+
+	if cfg.NetworkName != networkName {
```
abi87, line 207: this is one of the main addition of this PR.

service/backend/pchain/backend.go
```diff
@@ -25,7 +24,8 @@ var (
 
 type Backend struct {
 	fac                    *crypto.FactorySECP256K1R
-	networkIdentifier      *types.NetworkIdentifier
+	networkID              *types.NetworkIdentifier
+	networkHRP             string
```
abi87, line 28: `networkHRP` is one of those quantities depending only on networkID. Prefetching it upon Backend construction

cmd/server/main.go
```diff
@@ -196,6 +187,32 @@ func main() {
...
+		return fmt.Errorf("network name is not provided, can't fetch network name in offline mode")
+	}
+
+	log.Println("fetching network name from rpc...")
+	networkName, err := cChainClient.GetNetworkName(context.Background())
+	if err != nil {
+		return fmt.Errorf("cant fetch network name: %w", err)
+	}
+
+	if cfg.NetworkName == "" {
```
rajranjan0608, line 201: Should we use this the other way around, and put it before fetching the network name from the client?

```go
if cfg.NetworkName != "" {
  if cfg.NetworkName != networkName {
    // return error
  }

  return nil
}

// fetch from client
```

abi87: Indeed I was calling the backend client even in case service was offline. Fixed, not exactly as you suggested, but in an equivalent way

service/backend/pchain/backend.go
```diff
@@ -134,3 +141,14 @@ func (b *Backend) initChainIDs() error {
...
+// isPChain checks network identifier to make sure sub-network identifier set to "P"
+func (b *Backend) isPChain(reqNetworkID *types.NetworkIdentifier) bool {
+	return reqNetworkID != nil &&
+		reqNetworkID.SubNetworkIdentifier != nil &&
+		reqNetworkID.SubNetworkIdentifier.Network == mapper.PChainNetworkIdentifier
+}
+
+func (b *Backend) isNetworkSupported(reqNetworkID *types.NetworkIdentifier) bool {
+	return reqNetworkID != nil && reqNetworkID.Network == b.networkID.Network
+}
```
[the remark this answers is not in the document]
abi87: Great, rolling it back

PR #164, opened 2022-10-21
Pchain genesis repack

abi87:


service/backend/pchain/genesis_handler.go
```diff
@@ -0,0 +1,135 @@
...
+		}
+
+		out, ok := outIntf.(*secp256k1fx.TransferOutput)
+
+		if !ok {
+			return nil, errUnableToParseUTXO
+		}
+
+		outs = append(outs, &avax.TransferableOutput{
+			Out: &secp256k1fx.TransferOutput{
```
bilgehansahin, line 112: I found a bug here - we were not specifying the asset id. I opened #161  to fix, feel free to pull that in here as well. Then I can close #161 in favor of this.

abi87: merged

PR #165, opened 2022-10-21
pchain tx parser cleanup

abi87:


service/backend/pchain/block.go
```diff
@@ -30,97 +31,114 @@ func (b *Backend) Block(ctx context.Context, request *types.BlockRequest) (*type
...
+		metadata            map[string]interface{}
+	)
+
+	if b.isGenesisBlockRequest(blockIndex, hash) {
 		genesisTxs, err := b.getFullGenesisTxs()
 		if err != nil {
 			return nil, service.WrapError(service.ErrClientError, err)
 		}
-		rosettaTxs, err := b.parseRosettaTxs(genesisTxs, nil)
+		rosettaTxs, err := parseRosettaTxs(b.txParserCfg, blocks.GenesisCodec, genesisTxs, nil)
```
abi87, line 47: We should always use `blocks.GenesisCodec` instead of `blocks.Codec` to parse genesis and its transactions. Reason is we allow genesis content to be larger than the maximum size an ordinary block/transaction can have

service/backend/pchain/types.go
```diff
@@ -42,32 +46,26 @@ func (p *pTx) Unmarshal(bytes []byte) error {
 	if err != nil {
 		return err
 	}
+	if err := tx.Sign(p.Codec, nil); err != nil {
```
abi87, line 49: tx.Sign with nil signer carries out all the initialization a tx needs. I reckon this may not be super apparent from avalanchego codebase inspection. Still we should use it anywhere an initialization is needed

mapper/cchainatomictx/tx_parser.go
```diff
@@ -24,14 +25,14 @@ var (
 type TxParser struct {
 	// hrp used for address formatting
 	hrp string
-	// chainIDs contain chain id to chain id alias mappings
-	chainIDs map[string]string
+	// chainIDs maps chain id to chain id alias mappings
```
bilgehansahin, line 28: nit: chainIDs maps chain ids to chain id aliases

service/backend/pchain/block.go
bilgehansahin: Yeah, this got me before since testnet genesis is small enough to fit into the `blocks.Codec` and we only found out about this after trying to parse mainnet blocks 😄

PR #167, opened 2022-10-25
Accounts nits

abi87:


service/backend/pchain/account.go
```diff
@@ -117,38 +117,56 @@ func (b *Backend) AccountCoins(ctx context.Context, req *types.AccountCoinsReque
...
 			Index: int64(height),
 			Hash:  block.BlockID.String(),
 		},
-		Coins: common.SortUnique(coins),
+		Coins: coins,
 	}, nil
 }
 
-func (b *Backend) fetchBalance(ctx context.Context, addrString string, fetchImportable bool, currencyAssetIds map[ids.ID]struct{}) (uint64, *AccountBalance, *types.Error) {
+func (b *Backend) fetchBalance(ctx context.Context, addrString string, fetchImportable bool, assetIds ids.Set) (uint64, *AccountBalance, *types.Error) {
```
abi87, line 163: ids.Set is basically an alias for map[ids.ID]struct{} with some convenience methods along it.

service/backend/pchain/account.go
```diff
@@ -428,29 +446,3 @@ func (b *Backend) getAccountUTXOs(ctx context.Context, addr ids.ShortID, sourceC
 
 	return utxos, nil
 }
-
-func (b *Backend) processUtxos(currencyAssetIDs map[ids.ID]struct{}, utxos []avax.UTXO) ([]*types.Coin, error) {
```
abi87, line 432: I know there is a specular methods in cchainatomic backend, but I feel this method is not really doing that much and it does not really need to skip currencies since they are already filtered out. Absorbed this method into caller

Makefile
```diff
@@ -11,7 +11,7 @@ DOCKER_ORG          ?= avaplatform
 DOCKER_IMAGE        ?= ${DOCKER_ORG}/${PROJECT}
 DOCKER_LABEL        ?= latest
 DOCKER_TAG          ?= ${DOCKER_IMAGE}:${DOCKER_LABEL}
-AVALANCHE_VERSION   ?= v1.8.6
```
abi87, line 14: This diff is messed up. It happens with github at times, not sure how to fix it

rajranjan0608 (approved):

PR #172, opened 2022-10-26
TxDependency Bugfix

abi87:
Spotted by @bilgehansahin, fixed and added UT for it


mapper/pchain/tx_dependency.go
```diff
@@ -115,6 +115,7 @@ func (d *SingleTxDependency) GetUtxos() map[avax.UTXOID]*avax.UTXO {
 	if d.utxosMap != nil {
 		return d.utxosMap
 	}
+	d.utxosMap = make(map[avax.UTXOID]*avax.UTXO)
```
abi87, line 118: This was causing a crash

rajranjan0608 (approved):

PR #174, opened 2022-10-27
Network name-backward compatible fix

abi87:
Avalanchego codebase defines lower case network names ("mainnet", "fuji") while Rosetta uses are used to upper case ones.
This PR keep allowing user to use uppercase network identifiers and takes care of lowercasing it in case of user requests

### DIFF
--- a/cmd/server/config.go
+++ b/cmd/server/config.go
@@ -4,6 +4,7 @@ import (
 	"encoding/json"
 	"errors"
 	"os"
+	"strings"
 
 	ethcommon "github.com/ethereum/go-ethereum/common"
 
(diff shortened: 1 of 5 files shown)


cmd/server/config.go
```diff
@@ -72,6 +73,11 @@ func (c *config) applyDefaults() {
 }
 
 func (c *config) validate() error {
+	// avalanchego node define lowercase network names.
+	// We silently change config to ensure backward compatibility
+	// with previous deployments.
+	c.NetworkName = strings.ToLower(c.NetworkName)
```
bilgehansahin, line 79: I've just tested this. Rosetta asserter compares request network with what's provided to it at https://github.com/ava-labs/avalanche-rosetta/blob/pchain-support/cmd/server/main.go#L163-L166. As a result of this, capitalized network names in the requests still fail.

PR #176, opened 2022-10-28
[Codebase cleanup] Introducing CChain backend

abi87:


bilgehansahin:
You should update https://github.com/ava-labs/avalanche-rosetta/blob/pchain-support/.gitignore#L29 to include the test data json files in their new location

PR #177, opened 2022-10-28
Enable check:construction in Github actions for private network

xiaying-peng:
Enable rosetta-cli check:construction in CI.


xiaying-peng:
@patrick-ogrady Could you review it when you have time please?

PR #178, opened 2022-10-31
Enable check:construction in Github actions for private network

xiaying-peng:


xiaying-peng:
@patrick-ogrady Could you take a look please

PR #182, opened 2022-11-08
Fix Pchain Block ID usage

abi87:
Rosetta wants to use PChain block IDs as blocks identifiers and not proposerVM block IDs. Hence we want to forbid getting blocks by ID from indexer, which would assume the input hash is a proposerVM blockID and we use platformvm.GetBlock in such case

### DIFF
--- a/service/backend/pchain/indexer/parser.go
+++ b/service/backend/pchain/indexer/parser.go
@@ -83,7 +83,7 @@ func (p *parser) GetPlatformHeight(ctx context.Context) (uint64, error) {
 	if err != nil {
 		return 0, err
 	}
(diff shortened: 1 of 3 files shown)


service/backend/pchain/indexer/parser.go
```diff
@@ -184,20 +188,21 @@ func (p *parser) parseBlockWithHash(ctx context.Context, hash string) (*ParsedBl
 		return nil, err
 	}
 
-	container, err := p.pChainClient.GetContainerByID(ctx, hashID)
+	// Note that has
```
bilgehansahin, line 191: nit: comment

bilgehansahin (approved):

PR #188, opened 2022-12-13
Stricter Validity Checks + Nits

patrick-ogrady:


service/service_construction.go
```diff
@@ -111,6 +111,8 @@ func (s ConstructionService) ConstructionMetadata(
 			if input.Metadata != nil {
 				if input.Metadata.UnwrapBridgeTx {
 					gasLimit, err = s.getBridgeUnwrapTransferGasLimit(ctx, input.From, input.Value, input.Currency)
+				} else {
+					return nil, wrapError(errInvalidInput, "UnwrapBridgeTx must be populated if input.Metadata is provided")
```
patrick-ogrady, line 115: This would've caused the estimation to be 0 if there was metadata but no `UnwrapBridgeTx` boolean

service/service_construction.go
```diff
@@ -111,6 +111,8 @@ func (s ConstructionService) ConstructionMetadata(
...
 				}
```
dhrubabasu, line 116: I think this reads better:

```suggestion
				if !input.Metadata.UnwrapBridgeTx {
					return nil, wrapError(errInvalidInput, "UnwrapBridgeTx must be populated if input.Metadata is provided")
				}
				
				gasLimit, err = s.getBridgeUnwrapTransferGasLimit(ctx, input.From, input.Value, input.Currency)
```

patrick-ogrady: Agreed...was too confined to fitting this within the code.

service/service_construction.go
```diff
@@ -309,21 +311,28 @@ func (s ConstructionService) ConstructionParse(
...
 
+	var (
+		ops        []*types.Operation
+		checkFrom  *string
+		wrappedErr *types.Error
+	)
 	if len(tx.Data) != 0 {
 		unwrapMethodID := getMethodID(unwrapFnSignature)
+		transferMethodID := getMethodID(transferFnSignature)
 		if hexutil.Encode(tx.Data[:4]) == hexutil.Encode(unwrapMethodID) {
```
dhrubabasu, line 323: was gonna put a comment about the switch statement but looks like the linter caught it first haha

service/service_construction.go
dhrubabasu: 4580e42

service/service_construction.go
dhrubabasu: af10ff8

learyce (approved):

PR #190, opened 2022-12-27
no indexer call on genesis

abi87:
We initialize P-chain genesis block information on startup. However we should take care to do it in a backward-compatible manner, for installation whose avalanchego client does not have an indexer.
We achieve this by building genesis the way it's done in platformvm.
There is currently some code duplication across rosetta and avalanchego repo, which must be solved by making avalanchego expose facilities to build the genesis block.


service/backend/pchain/indexer/parser.go
```diff
@@ -110,14 +113,14 @@ func (p *parser) GetGenesisBlock(ctx context.Context) (*ParsedGenesisBlock, erro
...
 
-	// Genesis commit block's parent ID is the hash of genesis state
-	var genesisParentID ids.ID = hashing.ComputeHash256Array(bytes)
-
-	// Genesis Block is not indexed by the indexer, but its block ID can be accessed from block 0's parent id
-	genesisChildBlock, err := p.parseBlockAtHeight(ctx, 1)
+	// Build genesis ID and ParentID as it's done in platformVM'State,
+	// without polling indexer (for backward compatibility).
+	genesisParentID := hashing.ComputeHash256Array(bytes)
+	genesisBlock, err := pBlocks.NewApricotCommitBlock(genesisParentID, 0)
```
abi87, line 119: this is how it's done in platformVM. See https://github.com/ava-labs/avalanchego/blob/master/vms/platformvm/state/state.go#L1453:L1457

service/backend/pchain/indexer/testdata/outs/genesis_fuji_txs.json
```diff
@@ -1,5 +1,5 @@
 {
-  "id": "2FUFPVPxbTpKNn39moGSzsmGroYES4NZRdw3mJgNvMkMiMHJ9e",
+  "id": "99BWrAqUMvTp9nXKXyjPsCqjGwDqVFqssTRQbu58af57Cf9VG",
```
abi87, line 2: previous value was mainnet genesis ID, not fuji

PR #194, opened 2023-01-16
Get network name fix

abi87:
When running Rosetta we cannot take for granted that avalanchego node backing Rosetta server is ready when server is started.
I did not take this into account and mistakenly try and call GetNetworkName immediately upon startup.
This PR fixes the issue

### DIFF
--- a/client/client.go
+++ b/client/client.go
@@ -9,7 +9,6 @@ import (
 	"github.com/ava-labs/avalanchego/api"
 	"github.com/ava-labs/avalanchego/api/info"
 	"github.com/ava-labs/avalanchego/ids"
-	"github.com/ava-labs/avalanchego/utils/rpc"
 	ethtypes "github.com/ava-labs/coreth/core/types"
 	"github.com/ava-labs/coreth/interfaces"
 	"github.com/ava-labs/coreth/plugin/evm"
@@ -21,11 +20,7 @@ var _ Client = &client{}
 
 type Client interface {
 	// info.Client methods
-	IsBootstrapped(context.Context, string, ...rpc.Option) (bool, error)
-	GetNetworkName(context.Context, ...rpc.Option) (string, error)
-	Peers(context.Context, ...rpc.Option) ([]info.Peer, error)
-	GetNetworkID(context.Context, ...rpc.Option) (uint32, error)
-	GetBlockchainID(context.Context, string, ...rpc.Option) (ids.ID, error)
+	InfoClient
 
 	ChainID(context.Context) (*big.Int, error)
 	BlockByHash(context.Context, ethcommon.Hash) (*ethtypes.Block, error)
--- a/cmd/server/config.go
+++ b/cmd/server/config.go
@@ -78,6 +78,14 @@ func (c *config) validate() error {
 		return errInvalidMode
 	}
 
+	if c.Mode == service.ModeOffline && c.ChainID == 0 {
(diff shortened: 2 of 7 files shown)


client/client.go
```diff
@@ -21,11 +20,7 @@ var _ Client = &client{}
 
 type Client interface {
 	// info.Client methods
-	IsBootstrapped(context.Context, string, ...rpc.Option) (bool, error)
-	GetNetworkName(context.Context, ...rpc.Option) (string, error)
-	Peers(context.Context, ...rpc.Option) ([]info.Peer, error)
-	GetNetworkID(context.Context, ...rpc.Option) (uint32, error)
-	GetBlockchainID(context.Context, string, ...rpc.Option) (ids.ID, error)
+	InfoClient
```
abi87, line 23: just a minor consolidation, which I thought was necessary but turned out it really isn't. Still it should not be bad

service/backend/pchain/backend.go
```diff
@@ -67,24 +67,24 @@ func NewBackend(
...
-		if err := backEnd.initChainIDs(); err != nil {
+		if err := b.initChainIDs(); err != nil {
 			return nil, err
 		}
-	}
 
-	backEnd.networkHRP, err = mapper.GetHRP(networkIdentifier)
-	if err != nil {
-		return nil, err
+		if b.networkHRP, err = mapper.GetHRP(b.networkID); err != nil {
```
abi87, line 74: note: we only configure networkHRP correctly if in online mode. Remember that in validateNetworkName we make sure networkID is blank in Offline mode.

cmd/server/main.go
```diff
@@ -207,29 +206,15 @@ func validateNetworkName(cfg *config, cChainClient client.Client) error {
 		if cfg.NetworkName != "" {
 			return fmt.Errorf("network name is not provided, can't fetch network name in offline mode")
```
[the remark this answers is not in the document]
bilgehansahin: I don't think we can do that. We are expected to inject network name in https://github.com/ava-labs/avalanche-rosetta/blob/get_network_name_fix/cmd/server/main.go#L120-L130. 

In case of offline mode, I think we should trust the user to set the correct network name in the config.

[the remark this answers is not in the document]
abi87: So I understand you're say we always want NetworkName configured, whether it's online or offline mode.
As such I moved the network name validation to the cfg.validate method, where all "static" checks are carried out.

cmd/server/main.go
```diff
@@ -85,11 +83,6 @@ func main() {
 
 	if cfg.ChainID == 0 {
 		log.Println("chain id is not provided, fetching from rpc...")
-
-		if cfg.Mode == service.ModeOffline {
-			log.Fatal("cant fetch chain id in offline mode")
-		}
```
abi87, line 91: moved this check to cfg.validate method

cmd/server/config.go
```diff
@@ -78,6 +78,14 @@ func (c *config) validate() error {
 		return errInvalidMode
 	}
 
+	if c.Mode == service.ModeOffline && c.ChainID == 0 {
+		return errors.New("chainID must be configured when offline mode is selected")
+	}
+
+	if c.NetworkName == "" {
+		return errors.New("network name not provided")
+	}
```
abi87, line 87: network name must be specified for both online and offline mode

PR #195, opened 2023-01-18
Lazy inits

abi87:
Turns out it wasn't only GetNetworkName being called too early in some situations. Other avalanchego calls are issued upon server construction which causes the whole node to fail if avalanche node is not ready to serve requests immediately.
This PR addresses the issue

### DIFF
--- a/service/backend/pchain/backend.go
+++ b/service/backend/pchain/backend.go
@@ -1,8 +1,6 @@
 package pchain
 
 import (
-	"context"
-
 	"github.com/ava-labs/avalanchego/codec"
 	"github.com/ava-labs/avalanchego/ids"
 	"github.com/ava-labs/avalanchego/utils/crypto"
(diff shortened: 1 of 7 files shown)


service/backend/pchain/backend.go
```diff
@@ -79,7 +73,7 @@ func NewBackend(
 	b.txParserCfg = pmapper.TxParserConfig{
 		IsConstruction: false,
 		Hrp:            b.networkHRP,
-		ChainIDs:       b.chainIDs,
+		ChainIDs:       nil,
```
abi87, line 76: ChainIDs explicitly set to nil to lazy load them

service/backend/pchain/construction_test.go
```diff
@@ -874,8 +868,8 @@ func TestAddDelegatorTxConstruction(t *testing.T) {
 
 	ctx := context.Background()
 	clientMock := &mocks.PChainClient{}
-	clientMock.Mock.On("GetBlockchainID", ctx, constants.CChain.String()).Return(cChainID, nil)
-	clientMock.Mock.On("GetBlockchainID", ctx, constants.XChain.String()).Return(ids.ID{'X'}, nil)
+	// clientMock.Mock.On("GetBlockchainID", ctx, constants.CChain.String()).Return(cChainID, nil)
+	// clientMock.Mock.On("GetBlockchainID", ctx, constants.XChain.String()).Return(ids.ID{'X'}, nil)
```
bilgehansahin, line 872: nit: commented code

bilgehansahin (approved):

PR #197, opened 2023-02-06
disable state-sync by default

darioush:
Disable state sync by default on rosetta.

### DIFF
--- a/docker/entrypoint.sh
+++ b/docker/entrypoint.sh
@@ -23,6 +23,7 @@ mkdir -p /app/configs/chains/C
 
 cat <<EOF > /app/configs/chains/C/config.json
 {
+  "state-sync-enables": false,
   "snowman-api-enabled": false,
   "coreth-admin-api-enabled": false,
   "rpc-gas-cap": 2500000000,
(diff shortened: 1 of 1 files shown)


docker/entrypoint.sh
```diff
@@ -54,4 +55,4 @@ exec /app/rosetta-runner \
   -avalanche-bin /app/avalanchego \
   -avalanche-config /app/avalanchego-config.json \
   -rosetta-bin /app/rosetta-server \
-  -rosetta-config rosetta-config.json
\ No newline at end of file
+  -rosetta-config rosetta-config.json
```
aaronbuchwald, line 58: Nice fix for new line

docker/entrypoint.sh
```diff
@@ -23,6 +23,7 @@ mkdir -p /app/configs/chains/C
 
 cat <<EOF > /app/configs/chains/C/config.json
 {
+  "state-sync-enabled": false,
```
aaronbuchwald, line 26: This looks right to me to explicitly disable state sync

aaronbuchwald: For context, we are changing the default state sync behavior so that state sync is enabled by default for any node that is starting from scratch.

https://github.com/ava-labs/coreth-internal/pull/1178

aaronbuchwald (approved):
LGTM

learyce (approved):

PR #202, opened 2023-02-28
avalanchego-v1.9.9

rajranjan0608:


learyce (approved):

client/contract.go
```diff
@@ -13,14 +13,14 @@ const (
 // ContractClient is a client for the calling contract information
 type ContractClient struct {
 	ethClient ethclient.Client
-	cache     *cache.LRU
+	cache     *cache.LRU[common.Address, interface{}]
```
dhrubabasu, line 16: I think this just stores common.Address -> `*ContractInfo`

rajranjan0608: done. thanks

dhrubabasu (approved):

PR #223, opened 2023-09-21
Support arbitrary contract calls

cryptoriver:
### What
This pr enables arbitrary contract call support in avalanche rosetta implementation by introducing a utility function `constructContractCallDataGeneric` for encoding data field for any contract call method and its arguments.

Other than that, we updated the handlers of these endpoints to enable producing/carrying contract call data.

* /construction/preprocess
* /construction/metadata
* /construction/payloads
* /construction/parse

**Note:** If argument list contains tuple, the client side must provide the encoded ABI data as the only argument in the argument list because encoding tuple type in this rosetta implementation is not yet supported.

### Test
* Unit tests
* End to end tests for making generic contract call. ([explorer](https://testnet.snowtrace.io/tx/0xdde4e8adc834a097a85ad929c6587ff1952ee806531e6567bfec4af38d4a8fe6))
* For rosetta-cli test, the node sync will take too much time (more than 1 day) so decide to skip this (should be fine as the relevant use cases are covered in below end to end regression test).

#### regression test
* unwrap call ([explorer](https://testnet.snowtrace.io/tx/0x32d3502812c232cdfc1a36532da2e9d5745308867474499969cddd5492787588))
* avax transfer([explorer](https://testnet.snowtrace.io/tx/0xda5bb7e76049615417941902db60f578b989852880de57357228bbf187605c74))
* erc20 transfer ([explorer](https://testnet.snowtrace.io/tx/0x56eecc32b82a5b2be61263839ca3c75550c1b1339faca7842e5f8009e3c1e019))


Dockerfile.arm64
```diff
@@ -0,0 +1,71 @@
+# ------------------------------------------------------------------------------
+# Build avalanche
+# ------------------------------------------------------------------------------
+FROM arm64v8/golang:1.20.8-bullseye AS avalanche
```
cryptoriver, line 4: Add this docker file dedicated for build on Mac M1.

service/service_construction.go
```diff
@@ -577,6 +633,8 @@ func (s ConstructionService) ConstructionPayloads(
 
 	if isUnwrapRequest(req.Metadata) {
 		tx, unsignedTx, checkFrom, wrappedErr = s.createUnwrapPayload(req)
+	} else if isGenericContractCall(req.Metadata) {
+		tx, unsignedTx, checkFrom, wrappedErr = s.createGenericContractCallPayload(req)
```
cryptoriver, line 637: To my understanding, the `unwrap` is just another ordinary contract call so ideally, we may want to consolidate both into one. But in my opinion, I suggest to tackle that in a separate PR as the current rosetta operation definition for `unwrap` has major distinction(it has only 1 operation in the `ConstructionPreprocessRequest` rather than 2).

service/contract_call_data.go
```diff
@@ -0,0 +1,181 @@
...
+			}
+		case strings.HasPrefix(v, "bool"):
+			{
+				value, err := strconv.ParseBool(methodArgs[i])
+				if err != nil {
+					log.Fatal(err)
+				}
+				argData = value
+			}
+		}
```
ceyonur, line 158: should we add a `default:` case here?

service/contract_call_data.go
```diff
@@ -0,0 +1,181 @@
...
+// contractCallMethodID calculates the first 4 bytes of the method
+// signature for function call on contract
+func contractCallMethodID(methodSig string) ([]byte, error) {
+	fnSignature := []byte(methodSig)
+	hash := sha3.NewLegacyKeccak256()
+	if _, err := hash.Write(fnSignature); err != nil {
+		return nil, err
+	}
+
+	return hash.Sum(nil)[:4], nil
```
ceyonur, line 180: Can we replace this with go-ethereum's crypto library?
https://github.com/ava-labs/subnet-evm/blob/master/accounts/abi/method.go#L127-L128

Dockerfile.arm64
cryptoriver: correct, dedicated for building on Mac M1.

README.md
```diff
@@ -36,7 +36,7 @@ Before you start running the server you need to create a configuration file:
   "rpc_base_url": "https://api.avax-test.network",
   "mode": "online",
   "listen_addr": "0.0.0.0:8080",
-  "genesis_block_hash" :"0x31ced5b9beb7f8782b014660da0cb18cc409f121f408186886e1ca3e8eeca96b",
+  "genesis_block_hash" :"0x9e1c1fd401bbfdc58540b040fbdd5e91f916a080a855d01e9ea1bb6c74c15333",
```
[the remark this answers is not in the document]
cryptoriver: sure!

README.md
```diff
@@ -58,7 +58,7 @@ Full configuration example:
   "network_name": "Fuji",
   "chain_id": 43113,
   "log_requests": true,
-  "genesis_block_hash" :"0x31ced5b9beb7f8782b014660da0cb18cc409f121f408186886e1ca3e8eeca96b",
+  "genesis_block_hash" :"0x9e1c1fd401bbfdc58540b040fbdd5e91f916a080a855d01e9ea1bb6c74c15333",
```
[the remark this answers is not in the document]
cryptoriver: sure

docker/entrypoint.sh
```diff
@@ -3,7 +3,7 @@
 export AVALANCHE_NETWORK=${AVALANCHE_NETWORK:-testnet}
 export AVALANCHE_CHAIN=${AVALANCHE_CHAIN:-43113}
 export AVALANCHE_MODE=${AVALANCHE_MODE:-online}
-export AVALANCHE_GENESIS_HASH=${AVALANCHE_GENESIS_HASH:-"0x31ced5b9beb7f8782b014660da0cb18cc409f121f408186886e1ca3e8eeca96b"}
+export AVALANCHE_GENESIS_HASH=${AVALANCHE_GENESIS_HASH:-"0x9e1c1fd401bbfdc58540b040fbdd5e91f916a080a855d01e9ea1bb6c74c15333"}
```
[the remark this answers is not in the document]
cryptoriver: sure

service/contract_call_data.go
cryptoriver: sure, will add that

service/contract_call_data.go
cryptoriver: sure

service/types.go
```diff
@@ -106,30 +123,45 @@ func (o *options) UnmarshalJSON(data []byte) error {
...
-	GasPrice       *big.Int `json:"gas_price"`
-	GasLimit       uint64   `json:"gas_limit"`
-	UnwrapBridgeTx bool     `json:"bridge_unwrap"`
+	Nonce           uint64      `json:"nonce"`
+	GasPrice        *big.Int    `json:"gas_price"`
+	GasLimit        uint64      `json:"gas_limit"`
+	UnwrapBridgeTx  bool        `json:"bridge_unwrap"`
+	ContractData    string      `json:"data,omitempty"`
+	MethodSignature string      `json:"method_signature,omitempty"`
+	MethodArgs      interface{} `json:"method_args,omitempty"`
```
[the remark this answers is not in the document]
cryptoriver: Let's keep it as `interface{}` type. The `// case 2` in the `constructContractCallDataGeneric` method will convert each individual argument to string before encoding.

This is to follow the example rosetta implementation in the [rosetta-geth-sdk](https://github.com/coinbase/rosetta-geth-sdk/blob/master/client/types.go#L89) and also being applied by other node (e.g, [optimism rosetta](https://github.com/Inphi/optimism-rosetta/blob/master/services/types.go#L242)).

service/contract_call_data.go
```diff
@@ -0,0 +1,177 @@
+package service
+
+import (
+	"encoding/hex"
```
xiaoxiaff, line 4: small nit: grouping

service/service_construction.go
```diff
@@ -786,6 +844,85 @@ func (s ConstructionService) createUnwrapPayload(
...
+		OperationDescriptions: operationDescriptions,
+		ErrUnmatched:          true,
+	}
+
+	matches, err := parser.MatchOperations(descriptions, req.Operations)
+	if err != nil {
+		return nil, nil, nil, WrapError(ErrInvalidInput, "unclear intent")
+	}
+
+	fromOp, amount := matches[0].First()
```
xiaoxiaff, line 863: Can you add a check to make sure matches list indeed has two elements?

service/service_construction.go
```diff
@@ -1237,6 +1508,21 @@ func isUnwrapRequest(metadata map[string]interface{}) bool {
 	return false
 }
 
+func isGenericContractCall(metadata map[string]interface{}) bool {
+	if isUnwrap, ok := metadata["bridge_unwrap"]; ok {
+		if isUnwrap.(bool) {
```
xiaoxiaff, line 1513: Can you check the panic here?

service/service_construction_test.go
```diff
@@ -4,6 +4,7 @@ import (
 	"context"
 	"encoding/hex"
 	"encoding/json"
+	"github.com/ethereum/go-ethereum/common/hexutil"
```
xiaoxiaff, line 7: small nit: grouping

service/service_construction.go
cryptoriver: sure

service/contract_call_data.go
cryptoriver: sure

service/service_construction.go
cryptoriver: Do u mean checking `bridge_unwrap` not existing in the `metadata` map ? According to the `blockchain-avacchain` repo, this is a valid case as for native and token transfer, metadata map does not have this field.

Dockerfile.arm64
patrick-ogrady, line 4: The minimum supported version in AvalancheGo is 1.20.10: https://github.com/ava-labs/avalanchego/blob/2d8222da82114088a2a99d8d69f40846f1ea6208/scripts/build_avalanche.sh#L30

service/contract_call_data.go
```diff
@@ -0,0 +1,177 @@
...
+// It attempts to first convert the string arg to it's corresponding type in the method signature,
+// and then performs abi encoding to the converted args list and construct the data.
+func encodeMethodArgsStrings(methodID []byte, methodSig string, methodArgs []string) ([]byte, error) {
+	arguments := abi.Arguments{}
+	var argumentsData []interface{}
+
+	var data []byte
+	data = append(data, methodID...)
+
+	const split = 2
```
patrick-ogrady, line 79: nit: constants should be defined at the top of the file

service/contract_call_data.go
```diff
@@ -0,0 +1,177 @@
...
+	splitSigByLeadingParenthesis := strings.Split(methodSig, "(")
+	if len(splitSigByLeadingParenthesis) < split {
+		return data, nil
+	}
+	splitSigByTrailingParenthesis := strings.Split(splitSigByLeadingParenthesis[1], ")")
+	if len(splitSigByTrailingParenthesis) < 1 {
+		return data, nil
+	}
+	splitSigByComma := strings.Split(splitSigByTrailingParenthesis[0], ",")
+
```
patrick-ogrady, line 89: nit: inconsistent new lines next to `if` checks

service/contract_call_data.go
```diff
@@ -0,0 +1,177 @@
...
+		return nil, fmt.Errorf("failed to encode arguments: %w", err)
+	}
+
+	data = append(data, abiEncodeData...)
+	return data, nil
+}
+
+// contractCallMethodID calculates the first 4 bytes of the method
+// signature for function call on contract
+func contractCallMethodID(methodSig string) ([]byte, error) {
```
patrick-ogrady, line 175: Should probably check methodSig min length, otherwise the following line could panic.

service/types.go
```diff
@@ -106,30 +123,45 @@ func (o *options) UnmarshalJSON(data []byte) error {
...
 		o.Nonce = nonce
 	}
+	if len(ow.ContractData) > 0 {
+		o.ContractData = ow.ContractData
+	}
 
 	return nil
 }
 
 type metadata struct {
```
patrick-ogrady, line 133: I think we should break this struct up like:
```
type metadata struct {
	Nonce           uint64      `json:"nonce"`
	GasPrice        *big.Int    `json:"gas_price"`
	GasLimit        uint64      `json:"gas_limit"`

	UnwrapBridgeTx  bool        `json:"bridge_unwrap"`

	ContractData    string      `json:"data,omitempty"`
	MethodSignature string      `json:"method_signature,omitempty"`
	MethodArgs      interface{} `json:"method_args,omitempty"`
}
```

patrick-ogrady: (same with the others)

service/types.go
```diff
@@ -161,6 +195,10 @@ func (m *metadata) UnmarshalJSON(data []byte) error {
 	}
 	m.Nonce = nonce
 
+	if len(mw.ContractData) > 0 {
```
patrick-ogrady, line 198: Same here, you can just set it to whatever is there.

service/service_construction.go
```diff
@@ -1237,6 +1512,23 @@ func isUnwrapRequest(metadata map[string]interface{}) bool {
 	return false
 }
 
+func isGenericContractCall(metadata map[string]interface{}) bool {
+	if isUnwrap, ok := metadata["bridge_unwrap"]; ok {
+		if unwrapCall, isBool := isUnwrap.(bool); isBool {
+			return !unwrapCall
```
patrick-ogrady, line 1518: Will `unwrapCall` ever be false here? If it is, it seems like we'll return early without checking `method_signature`...

patrick-ogrady: I think we only want to return false here if `unwrapCall` is true, otherwise we should continue.

service/service_construction.go
```diff
@@ -545,6 +552,55 @@ func createUnwrapOps(tx transaction) ([]*types.Operation, *string, *types.Error)
...
+		)
+	}
+
+	// Ensure valid to address
+	checkTo, ok := ChecksumAddress(tx.To)
+	if !ok {
+		return nil, nil, WrapError(ErrInvalidInput, fmt.Errorf("%s is not a valid address", tx.To))
+	}
+
+	ops := []*types.Operation{
```
patrick-ogrady, line 573: It seems kind of weird to me that the `data` for the call isn't in `operation.Metadata`?

https://www.rosetta-api.org/docs/models/Operation.html

patrick-ogrady: is this just because of the way you are indexing these operations?

README.md
```diff
@@ -133,6 +133,11 @@ Before we can start the service, we need to build the docker image:
 make docker-build
 ```
 
+**Note:** If you're running arm64 machine(e.g, Mac M1), you may run
+
+```bash
+make docker-build-arm64
+```
```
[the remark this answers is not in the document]
cryptoriver: thanks!

service/contract_call_data.go
```diff
@@ -0,0 +1,177 @@
...
+				}
+				copy(value[:], bytes)
+				argData = value
+			}
+		case strings.HasPrefix(v, "bytes"):
+			{
+				// No fixed size set as it would make it an "array" instead
+				// of a "slice" when encoding. We want it to be a slice.
+				value := []byte{}
+				bytes, err := hexutil.Decode(methodArgs[i])
```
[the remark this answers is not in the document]
cryptoriver: I got this code from other rosetta implementation. Agree we don't need this intermediate `value` slice and can directly assign the `bytes` to `argData`.

service/service_construction.go
cryptoriver: for native and token send transaction, there won't be `bridge_unwrap` field in the metadata. In such case, the `ok` will always be false and it will hit the Line 1529 to return false. The `method_signature` length check is being used as a extra safeguard there that if we don't set `bridge_unwrap` but method_signature.length > 0, we still treat it as generic contract call.

service/service_construction.go
cryptoriver: yeah, we don't put contract call data into the operation object, we put it in the `options` struct which is to follow the existing design like in this [rosetta geth sdk implementation](https://github.com/coinbase/rosetta-geth-sdk/blob/master/services/construction/preprocess.go#L155C11-L155C23)

service/types.go
```diff
@@ -106,30 +123,45 @@ func (o *options) UnmarshalJSON(data []byte) error {
...
-		UnwrapBridgeTx: m.UnwrapBridgeTx,
+		Nonce:           hexutil.Uint64(m.Nonce).String(),
+		GasPrice:        hexutil.EncodeBig(m.GasPrice),
+		GasLimit:        hexutil.Uint64(m.GasLimit).String(),
+		UnwrapBridgeTx:  m.UnwrapBridgeTx,
+		MethodSignature: m.MethodSignature,
+		MethodArgs:      m.MethodArgs,
+	}
+
+	if len(m.ContractData) > 0 {
```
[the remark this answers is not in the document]
cryptoriver: don't need it, will update.

service/types.go
cryptoriver: sure!

service/contract_call_data.go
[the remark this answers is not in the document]
cryptoriver: this code is getting from the [rosetta-geth-sdk](https://github.com/coinbase/rosetta-geth-sdk/blob/master/services/construction/contract_call_data.go#L148), I admit that it's not necessary, will update.

service/service_construction.go
patrick-ogrady: Got it. I think that gets away a bit from the intent of `Operations` (which should have full-fidelity of on-chain interaction) but that's not a convo for this forum.

service/service_construction.go
patrick-ogrady: I guess what I'm saying here is that if you set `bridge_unwrap` to false, it won't actually end up checking the value of any other metadata (with how it is currently written).

service/contract_call_data.go
```diff
@@ -173,5 +169,8 @@ func encodeMethodArgsStrings(methodID []byte, methodSig string, methodArgs []str
 // contractCallMethodID calculates the first 4 bytes of the method
 // signature for function call on contract
 func contractCallMethodID(methodSig string) ([]byte, error) {
+	if len(methodSig) == 0 {
```
patrick-ogrady, line 172: I think you need this to be < 4 based on your access below?

service/service_construction.go
cryptoriver: I see your point, we set `bridge_unwrap=false` only if this is a generic contract call on client side. For native and token send, this field `bridge_unwrap` won't exists.

But I understand it make more sense to not only check `bridge_unwrap` but also the method signature length it it is going to return true.

Will enhance this logic.

service/contract_call_data.go
cryptoriver: This `crypto.Keccak256([]byte(methodSig))[:4]`  is just to take the first 4 bytes of the Keccak-256 hash rather than the first 4 bytes of the method signature string. 

So checking the string not empty should be enough.

service/service_construction.go
```diff
@@ -1515,7 +1515,11 @@ func isUnwrapRequest(metadata map[string]interface{}) bool {
 func isGenericContractCall(metadata map[string]interface{}) bool {
 	if isUnwrap, ok := metadata["bridge_unwrap"]; ok {
 		if unwrapCall, isBool := isUnwrap.(bool); isBool {
-			return !unwrapCall
+			if unwrapCall {
+				// If bridge_unwrap flag is true, we can return false right away, otherwise we should
+				// continue to check the method signature length.
+				return false
```
patrick-ogrady, line 1521: Nice 👍

cryptoriver: Thank you!

PR #231, opened 2024-01-23
Update avalanchego version to v1.10.18

StephenButtolph:


.github/workflows/ci.yml
```diff
@@ -7,7 +7,7 @@ on:
   pull_request:
 
 env:
-  go_version: '~1.20.8'
```
StephenButtolph, line 10: This probably should have been updated when we updated the docker to `1.20.10`

dhrubabasu (approved):

PR #234, opened 2024-02-07
Use `mockgen` for all mocks

dhrubabasu:
Ports over the mock infrastructure used in https://github.com/ava-labs/avalanchego for easier maintenance

Removed all unused mocks as well


service/service_construction_test.go
```diff
@@ -906,15 +884,6 @@ func TestPreprocessMetadata(t *testing.T) {
...
 		}
-		currency := &types.Currency{Symbol: defaultSymbol, Decimals: defaultDecimals}
-		client.On(
-			"ContractInfo",
-			common.HexToAddress(defaultContractAddress),
-			true,
-		).Return(
-			currency,
-			nil,
-		).Once()
```
dhrubabasu, line 917: This was never being hit, was it supposed to be? cc @learyce

service/service_construction_test.go
```diff
@@ -1004,15 +970,6 @@ func TestPreprocessMetadata(t *testing.T) {
...
```
dhrubabasu, line 1015: This was never being hit, was it supposed to be? cc @learyce

service/backend/cchainatomictx/construction_test.go
```diff
@@ -167,8 +169,6 @@ func TestExportTxConstruction(t *testing.T) {
 
 		assert.Nil(t, apiErr)
 		assert.Equal(t, metadataOptions, resp.Options)
-
-		clientMock.AssertExpectations(t)
```
patrick-ogrady, line 171: Is there an equivalent asset that needs to be called?

patrick-ogrady (approved):

PR #236, opened 2024-02-09
Remove `pmapper.StakingOptions`

dhrubabasu:
This was a copy of `pmapper.StakingMetadata`. We can unmarshal into it directly


patrick-ogrady (approved):

patrick-ogrady (changes requested):
wait

PR #248, opened 2024-11-24
Rosetta Etna upgrade - Dynamic fee suggestions

rajranjan0608:
Right now, Avalanche Rosetta is capable of building Atomic and Primary Network Staking Txs only. Other subnet-related txs are not supported, so we don't need to include any changes related to the new post-Etna L1 validator tx types.

This PR just updates the flow to suggest dynamic fees based on transaction size and network activity.

- Upgrades AvalancheGo to Etna version - `v1.12.0-fuji`
- Updates `/preprocess` endpoint to calculate and return `base_fee` as an `Options` property based on the operations passed (inputs and outputs)
- Updates `/metadata` endpoint to calculate additional suggested fees based on the staking metadata


michaelkaplan13 (commented):
Should we bump the AvalancheGo version as last done in this PR for `v1.11.6`?
https://github.com/ava-labs/avalanche-rosetta/pull/246/files

mapper/pchain/tx_dependency.go
```diff
@@ -167,6 +177,16 @@ func (d *SingleTxDependency) GetUtxos() map[avax.UTXOID]*avax.UTXO {
 			outsToAdd = append(outsToAdd, unsignedTx.Stake()...)
 		case *txs.TransferSubnetOwnershipTx:
 			outsToAdd = append(outsToAdd, unsignedTx.Outputs()...)
+		case *txs.ConvertSubnetToL1Tx:
```
michaelkaplan13, line 180: Is it easy/worth to add unit tests for these cases?

service/backend/pchain/construction.go
```diff
@@ -45,7 +48,25 @@ func (*Backend) ConstructionPreprocess(
...
+		pmapper.OpDummyBase,
+		matches,
+		pmapper.Metadata{BlockchainID: ids.Empty},
+		b.codec,
+		b.avaxAssetID,
+	)
+	if err != nil {
+		return nil, service.WrapError(service.ErrInternalError, err)
+	}
+	fee, err := b.calculateFee(ctx, dummyBaseTx, 0)
```
bilgehansahin, line 63: Preprocessing is an offline endpoint, you cannot talk to the node here. You can only do that in Metadata and Submit. 

I'd recommend passing matches in options and calculating the fee there. This way you should be able to implement this once instead of splitting between import/export and staking cases.

client/pchainclient.go
```diff
@@ -61,6 +63,8 @@ type PChainClient interface {
 	IssueTx(ctx context.Context, tx []byte, options ...rpc.Option) (ids.ID, error)
 	GetStake(ctx context.Context, addrs []ids.ShortID, validatorsOnly bool, options ...rpc.Option) (map[ids.ID]uint64, [][]byte, error)
 	GetCurrentValidators(ctx context.Context, subnetID ids.ID, nodeIDs []ids.NodeID, options ...rpc.Option) ([]platformvm.ClientPermissionlessValidator, error)
+	GetFeeConfig(ctx context.Context, options ...rpc.Option) (*gas.Config, error)
```
bilgehansahin, line 66: Is this used anywhere?

mapper/pchain/tx_dependency.go
rajranjan0608: We are doing the same operations for all the tx types which is to add outputs to an array. I don't think testing the new transaction types will make any difference here.

service/backend/pchain/construction.go
rajranjan0608: Done

service/backend/pchain/construction.go
```diff
@@ -62,15 +67,37 @@ func (b *Backend) ConstructionMetadata(
 		return nil, service.WrapError(service.ErrInvalidInput, err)
 	}
 
-	var suggestedFee *types.Amount
+	if opMetadata.Matches == nil {
+		return nil, service.WrapError(service.ErrInvalidInput, errors.New("matches not found in options"))
+	}
+
+	// Build a dummy base tx to calculate the base fee
+	dummyBaseTx, _, err := pmapper.BuildTx(
```
bilgehansahin, line 75: Why do we have to use a dummy base tx instead of the actual type? Can we simply use opMetadata.Type for type and move this whole block after the switch case below, using the actual metadata?

This way you wouldn't have to buildTX twice for staking transactions for example.

bilgehansahin (approved):
Looks good to me.

go.mod
```diff
@@ -1,16 +1,18 @@
 module github.com/ava-labs/avalanche-rosetta
 
-go 1.21.12
+go 1.22.8
```
michaelkaplan13, line 3: Nit should we go right to `go 1.22.10` since it was recently released?

mapper/pchain/tx_parser.go
```diff
@@ -730,6 +765,13 @@ func (t *TxParser) isMultisig(utxoid avax.UTXOID) (bool, error) {
 	utxoMap := dependencyTx.GetUtxos()
 	utxo, ok := utxoMap[utxoid]
 	if !ok {
+		// Consider the UTXO as multisig if it is not found in the dependency tx.
+		// Possible for dependency txs like [DisableL1ValidatorTx] or [SetL1ValidatorWeightTx].
+		// These unfound UTXOs are generated by the network but are not included in the tx bytes.
+		switch dependencyTx.Tx.Unsigned.(type) {
+		case *txs.DisableL1ValidatorTx, *txs.SetL1ValidatorWeightTx:
+			return true, nil
+		}
```
michaelkaplan13, line 774: This case is for when the UTXO ID refers to a UTXO that was created by a `DisableL1ValidatorTx` or `SetL1ValidatorWeightTx` (where the weight is 0 such that its ending the validation)? In those cases, the UTXO is defined by the `RemainingBalanceOwner` specified in the ICM message of the `ReigsterL1ValidatorTx` that added the validator. 

Does it matter if the UTXO is actually a multisig or not in this case though? We theoretically could do the same check on the length of `Addresses` from that message if so. I suspect in most cases it will not be a multisig

rajranjan0608: Rosetta doesn't track multi-sig UTXOs. This case is added to skip the "CHANGE" utxos generated by `DisableL1ValidatorTx` or `SetL1ValidatorWeightTx`. Since we can't track the generation of these UTXOs (as these are not present transaction bytes), we skip them similar to multi-sig UTXOs.

michaelkaplan13 (approved):

PR #250, opened 2025-10-27
Granite upgrade - coreth/libevm usage

rajranjan0608:
Upgrades AvalancheGo (and dependent packages like Coreth and libevm) from 1.12.0 to 1.14.0


mapper/block.go
```diff
@@ -17,11 +18,14 @@ func BlockMetadata(block *types.Block) map[string]interface{} {
 	if block.BaseFee() != nil {
 		meta["base_fee"] = hexutil.EncodeBig(block.BaseFee())
 	}
-	if block.BlockGasCost() != nil {
-		meta["block_gas_cost"] = hexutil.EncodeBig(block.BlockGasCost())
+
```
ARR4N, line 21: You need to be using [emulation](https://github.com/ava-labs/avalanchego/pull/4430) here otherwise the `customtypes` package functions will panic.

That PR should be merged very soon, but in the meantime just look inside it to see the single function call made by `emulate.CChain()`.

If you're already wrapping at a higher level then it's ok and MUST NOT be done twice as nested calls to `WithTempRegisteredLibEVMExtras()` will deadlock as they attempt to hold the same `sync.Mutex`.

mapper/transaction.go
```diff
@@ -333,12 +334,12 @@ func CrossChainTransactions(
 ) ([]*types.Transaction, error) {
 	transactions := []*types.Transaction{}
 
-	extra := block.ExtData()
+	extra := customtypes.BlockExtData(block)
```
ARR4N, line 337: As above re emulation.

mapper/block.go
rajranjan0608: Using `emulate.CChain` now 👍

service/service_block.go
```diff
@@ -101,45 +101,69 @@ func (s *BlockService) Block(
...
-	return &types.BlockResponse{
-		Block: &types.Block{
-			BlockIdentifier:       blockIdentifier,
-			ParentBlockIdentifier: parentBlockIdentifier,
-			Timestamp:             int64(block.Time() * utils.MillisecondsInSecond),
-			Transactions:          append(transactions, crosstx...),
-			Metadata:              mapper.BlockMetadata(block),
-		},
-	}, nil
+	return blockResponse, nil
```
emilypi-avalabs: A couple of nits generally due to the fact that we're munging too many types of errors into the same `err` value and it ends up being wonky and overcomplicated: 

1. Rosetta errors should probably be made a distinct workflow here, separate from the block response value prep. 
2. Block response construction is arguably the point of this function, and prioritizing error parsing and focusing on that as the main work body makes the block response a side effect of error detection, which seems backwards to me. 
3.  As a result, a lot of things are more mutable than they need to be. For example, you could just construct `blockIdentifier` directly and `parentBlockIdentifier` directly too (with a little error handling). You also then do not need to declare the `blockResponse` var until you've set up the Rosetta txs and then you can just construct it directly. 
4. Variable shadowing is a personal peeve of mine. It's easier to read what's going on if you just name the `rosettaErr`s distinctly.

rajranjan0608: @emilypi-avalabs Thanks for the suggestions! I've removed the code changes here in favour of using a global type registration instead of nesting everything inside the emulate function https://github.com/ava-labs/avalanche-rosetta/pull/250/commits/259e632108fff492755e8062d96087d3c958868a

artemanosov (approved):

emilypi-avalabs (approved):
Nice job - much cleaner thanks

camharris (approved):
LGTM 

Thank you for your work on this 🙏🏿

PR #254, opened 2026-07-17
Helicon Update

artemanosov:


client/client.go
```diff
@@ -45,7 +45,7 @@ type Client interface {
 	EstimateBaseFee(ctx context.Context) (*big.Int, error)
 }
 
-type EvmClient evm.Client
+type EvmClient struct{ *evm.Client }
```
SayanKar, line 48: WHy are we wrapping this, in a struct?

artemanosov: because after update evm.Client is used as a pointer and later in the code `EvmClient` is embedded into `client` struct. You won't be able to do this embedding if it was just a pointer

mapper/pchain/tx_builder.go
```diff
@@ -576,6 +580,140 @@ func buildOutputs(
...
+) (*txs.Tx, []*types.AccountIdentifier, error) {
+	if metadata.AutoRenewedValidator == nil {
+		return nil, nil, errInvalidMetadata
+	}
+	m := metadata.AutoRenewedValidator
+
+	// Validate required address fields before the more expensive BLS parsing.
+	if len(m.ValidationRewardsOwners) == 0 {
+		return nil, nil, errors.New("reward_addresses must be non-empty for AddAutoRenewedValidatorTx")
+	}
```
emilypi-avalabs: Are these not gauranteed? seems like if we get metadata at this point, it should be parsed to be correct by construction and we can assume these predicates. (See below as well)

mapper/pchain/types.go
```diff
@@ -27,6 +27,11 @@ const (
 	OpSetL1ValidatorWeightTx       = "SET_L1_VALIDATOR_WEIGHT_TX"
 	OpDisableL1ValidatorTx         = "DISABLE_L1_VALIDATOR_TX"
 
+	// Helicon (ACP-236) transaction types
+	OpAddAutoRenewedValidator       = "ADD_AUTO_RENEWED_VALIDATOR"
```
emilypi-avalabs, line 31: Why is this fundamentally different than the previous validator?

artemanosov: It maps to a distinct avalanchego tx type (`AddAutoRenewedValidatorTx`), not a flavor of the permissionless one. There main difference is that there is no start/end window, it stakes for a recurring period and auto-renews.

mapper/pchain/tx_builder.go
artemanosov: Not really, but I will drop the check to be aligned with other functions

mapper/pchain/tx_builder.go
```diff
@@ -576,6 +580,135 @@ func buildOutputs(
...
+	m := metadata.AutoRenewedValidator
+
+	// reward_addresses is validated to be non-empty in buildAutoRenewedValidatorMetadata,
+	// so it is safe to fall back to it for the authority owner here.
+	authorityAddrs := m.ValidatorAuthorityOwners
+	if len(authorityAddrs) == 0 {
+		authorityAddrs = m.ValidationRewardsOwners
+	}
+
```
jasonatran: This seems like a weird fallback to use rewards owners. Is this correct? If not, let's have ValidatorAuthorityOwners be a required input.

jasonatran (approved):

service/backend/pchain/block.go
```diff
@@ -171,7 +171,24 @@ type depRequest struct {
 func (b *Backend) fetchBlkDependencies(ctx context.Context, blkTxs []*txs.Tx) (pmapper.BlockTxDependencies, error) {
 	blockDeps := make(pmapper.BlockTxDependencies)
 
-	var depReqs []depRequest
+	// depReqByTxID dedupes dependency requests by depTxID. Several txs in a block can
```
jasonatran, line 174: lgtm

service/backend/pchain/construction_test.go
```diff
@@ -1357,6 +1357,17 @@ func TestSetAutoRenewedValidatorConfigTxConstruction(t *testing.T) { //nolint:go
 		parsedPTx, ok := rosettaTx.Tx.(*pTx)
 		require.True(t, ok)
 		require.Len(t, parsedPTx.Tx.Creds, 2)
+
+		// /construction/parse of the signed tx must report both signers: the BaseTx
+		// input and the ValidatorAuthority auth signer. Before the ExtraSigners fix
```
jasonatran: nit: remove comment about ExtraSigners fix.

service/backend/pchain/construction.go
```diff
@@ -187,32 +187,44 @@ func (*Backend) buildAutoRenewedValidatorMetadata(
 	}
 
 	// buildOutputOwner returns an empty (unspendable) owner for an empty address
-	// list without erroring, so guard it here where the metadata is constructed.
+	// list without erroring, so guard both owner lists here where the metadata is
+	// constructed. validator_authority_addresses must be explicit: silently falling
+	// back to the reward addresses would grant the reward key permanent authority to
+	// reconfigure or stop the validator, which the caller never asked for.
```
jasonatran: nit: remove comment about the change for validator_authority_addresses

service/backend/pchain/construction.go
```diff
@@ -238,10 +238,19 @@ func (*Backend) buildAutoRenewedValidatorConfigMetadata(
 		return nil, err
 	}
 
-	// The authority key produces an extra signing payload in the construction
-	// flow, so its address is required.
-	if opts.AuthAddress == "" {
-		return nil, errors.New("auth_address must be non-empty")
+	// Each authority key produces an extra signing payload in the construction flow,
+	// so at least one is required. AuthAddresses[i] pairs with AuthSigIndices[i], and
+	// the indices must be strictly ascending, as secp256k1fx.Input requires.
```
jasonatran: nit: remove mention of secp256k1fx as it's not directly protocol related (define protocol as what's known from the interface level, and not necessarily the nuances of avalanchego code). I could be wrong on this though!

jasonatran (approved):

service/backend/pchain/construction_test.go
artemanosov: addressed

service/backend/pchain/construction.go
artemanosov: addressed

service/backend/pchain/construction.go
artemanosov: addressed

